PR #2, opened 2019-05-08
Inital linux loader crate

bjzhjing:
- Based on Firecracker and crosvm code
- Added bzImage support
- Added buildkite CI and kcov based code coverage


src/loader/mod.rs
```diff
@@ -0,0 +1,578 @@
...
+        v
+    }
+
+    #[allow(safe_packed_borrows)]
+    #[allow(non_snake_case)]
+    #[test]
+    #[ignore]
+    fn load_bzImage() {
+        let gm = create_guest_mem();
+        let image = make_bzImage();
```
andreeaflorescu: Even if you specify ignore, the test will still be compiled. Because of that compilation of tests fails with:
```
error: couldn't read src/loader/bzImage: No such file or directory (os error 2)
   --> src/loader/mod.rs:369:29
    |
369 |         v.extend_from_slice(include_bytes!("bzImage"));
    |                             ^^^^^^^^^^^^^^^^^^^^^^^^^

error: aborting due to previous error

error: Could not compile `linux-loader`.
```

I created an empty bzImage file to get rid of the error, but once we enable the CI it will fail there as well as the bzImage file doesn't exist. Right now I can't think of a nice way to go around this problem. There would be two not so nice options:
- add an empty bzImage file to the repository (I don't really like this one)
- make bzImage support a feature not enabled by default. This means that the tests will not be compiled by default either.

In the long run though I would like bzImage support to be tested as part of the CI as well to make sure we aren't introducing any bugs there. From a chat with @sameo I understood that we don't want to have a bzimage in the repository because of the license. How is this different from the elf image? Also, does the license apply to test code that is not part of the release binaries? It sounds kinda strange to me.

rbradford: @andreeaflorescu a bzImage would be compiled from the Linux kernel which is GPL-2 and so we would have to ship all the corresponding source / make an offer to provide the source.

The CI script for this repository could either fetch a prebuilt binary from an external source or could build a Linux kernel from scratch.

.buildkite/pipeline.yml
```diff
@@ -0,0 +1,61 @@
...
+    agents:
+      platform: x86_64.metal
+    plugins:
+      - docker#v3.0.1:
+          image: "fandree/rust-vmm-dev"
+          always-pull: true
+  
+  - label: "unittests-gnu-x86"
+    commands:
+     - cargo test
```
rbradford: In particular this is where the bzImage could be prepared.

src/loader/mod.rs
andreeaflorescu: I was thinking about providing an already built bzImage image like we are doing with elf. I don't see the need to always compile it from the linux source just so we can use it in the unit tests. For extensive testing with different kernel versions  we might want to result to integration testing instead.

rbradford: > I was thinking about providing an already built bzImage image like we are doing with elf. I don't see the need to always compile it from the linux source just so we can use it in the unit tests. For extensive testing with different kernel versions we might want to result to integration testing instead.

Unless we create something that looks like the bzImage, with it's custom binary layout, which would be a lot of work, the only straightforward way to get a bzImage is to build a Linux kernel. Shipping that built kernel in the repository would mean satisfying the GPL requirements which would be a large burden.

bjzhjing: @rbradford @andreeaflorescu Just want to confirm, to fix this issue, bzImage should be added to the server running CI test for rust-vmm crate, and specified under .buildkite/pipeline.yml as Rob mentioned. if my understanding is correct, will Andrea help put it or should I do?

.buildkite/pipeline.yml
bjzhjing: @rbradford @andreeaflorescu Just want to confirm, to fix this issue, bzImage should be added to the server running CI test for rust-vmm crate, and specified under .buildkite/pipeline.yml as Rob mentioned. if my understanding is correct, will Andrea help put it or should I do?

andreeaflorescu: @bjzhjing I cannot help you cause you misspelled my name :cry:

andreeaflorescu: I was thinking about asking around if someone knows a better way to use a bzImage binary in tests. It sounds absurd to be so complicated and I am sure that other people had this problem as well.

In any case, if there is no other way but having it build as a pre-step on the buildkite agent, I can work with you to make it happened. I think though that whoever writes the code also owns the testing, so I can provide info and access to the machine that is running the CI and you can do your own thing there. Would that be ok?

rbradford: I don't think the binary should be manually place on the build machine as then it makes it hard to reproduce. A simple script that either downloads a prebuilt kernel rpm/deb and extracts the kernel should be added to the repository and run as a step prior to the unit test running itself. This also means all CI configuration is stored in revision control.

In the meantime I propose that we merge a version without bzimage support and spin the commit adding that into it's own PR.

bjzhjing: @andreeaflorescu Hi Andreea, sorry for the typo of your name 

> @bjzhjing I cannot help you cause you misspelled my name 😢

Hi Andreea, I'm really sorry for making the typo, please forgive me! :-P

bjzhjing: @rbradford @andreeaflorescu I'm considering to add feature for Elf and bzImage to separate the test by define cargo test --features in the yaml file. If you both agree with that, I will make the change. 

And then, I think I need to get an account of buildkite, I'd like to see how to define an upload action for a pre-compiled bzImage in the yaml, and download it to linux-loader crate for test running. How do you think?

andreeaflorescu: @bjzhjing I was trolling you with the name, sorry! Sounds good to me to have it as a separate feature. For the buildkite account, please ping on email/slack and I'll give you access to the rust-vmm one.

tests/coverage
```diff
@@ -1 +1 @@
-63.0
\ No newline at end of file
+67.0
```
rbradford: Please swap the word order in your commit message, adjective before noun, i.e. "coverage: update for current higher coverage"

Cargo.toml
```diff
@@ -5,6 +5,12 @@ authors = ["cathy <cathy.zhang@intel.com>"]
 edition = "2018"
 license = "Apache-2.0 AND BSD-3-Clause"
 
+[features]
+default = []
+Elf = []
```
rbradford: All the example features i've seen are all in lowercase.

Cargo.toml
```diff
@@ -0,0 +1,16 @@
+[package]
+name = "linux-loader"
+version = "0.1.0"
+authors = ["cathy <cathy.zhang@intel.com>"]
+edition = "2018"
+license = "Apache-2.0 AND BSD-3-Clause"
+
+[features]
+default = []
```
rbradford: I think elf should be on by default.

src/loader/mod.rs
```diff
@@ -0,0 +1,580 @@
...
+
+use std::error::{self, Error as KernelLoaderError};
+use std::ffi::CStr;
+use std::fmt::{self, Display};
+use std::io::{Read, Seek, SeekFrom};
+use std::mem;
+
+use vm_memory::{Address, Bytes, GuestAddress, GuestMemory, GuestMemoryMmap, GuestUsize};
+
+#[allow(dead_code)]
```
andreeaflorescu: Why is this necessary?

.buildkite/pipeline.yml
```diff
@@ -0,0 +1,61 @@
+steps:
```
andreeaflorescu: Since we can have multiple builds for this crate we should also test all of those builds with the pipelines.
What I am talking about is having "cargo build --feature x", "cargo build --feature y" as well as the default build. Same for cargo test. We should also check whether `cargo kcov --all` actually runs all unit tests (including the one labeled with ignored. Ideally these tests should be in a different pipeline and the default pipeline would test only "cargo test" and "cargo build" and so on.

src/loader/mod.rs
```diff
@@ -0,0 +1,580 @@
...
+    fn make_elf_bin() -> Vec<u8> {
+        let mut v = Vec::new();
+        v.extend_from_slice(include_bytes!("test_elf.bin"));
+        v
+    }
+
+    #[allow(safe_packed_borrows)]
+    #[allow(non_snake_case)]
+    #[test]
+    #[ignore]
```
andreeaflorescu: Should this still be ignored? Do we have a workaround for it now?

src/loader/mod.rs
```diff
@@ -0,0 +1,580 @@
...
+    /// Loads a kernel from a vmlinux elf image to a slice
+    ///
+    /// kernel is loaded into guest memory at offset phdr.p_paddr specified by elf image.
+    ///
+    /// # Arguments
+    ///
+    /// * `guest_mem` - The guest memory region the kernel is written to.
+    /// * `kernel_start` - The offset into 'guest _mem' at which to load the kernel.
+    /// * `kernel_image` - Input vmlinux image.
+    /// * `lowest_kernel_start` - This is the start of the high memory, kernel should above it.
```
andreeaflorescu: Shouldn't this be something like `highmem_start_address` or just `start_address` since it is sort of implied that the kernel is loaded in the high memory? I personally prefer have the name of the variables suffixed with what they actually represent when it is not clear from the name (in this case I would use address).

src/loader/mod.rs
```diff
@@ -0,0 +1,580 @@
...
+
+impl KernelLoader for Elf {
+    /// Loads a kernel from a vmlinux elf image to a slice
+    ///
+    /// kernel is loaded into guest memory at offset phdr.p_paddr specified by elf image.
+    ///
+    /// # Arguments
+    ///
+    /// * `guest_mem` - The guest memory region the kernel is written to.
+    /// * `kernel_start` - The offset into 'guest _mem' at which to load the kernel.
```
andreeaflorescu: nit: extra space in 'guest _mem'

src/loader/mod.rs
```diff
@@ -0,0 +1,580 @@
...
+    /// * `kernel_image` - Input vmlinux image.
+    /// * `lowest_kernel_start` - This is the start of the high memory, kernel should above it.
+    ///
+    /// # Returns
+    /// * KernelLoaderResult
```
andreeaflorescu: nit: `KernelLoaderResult` to keep the same style as above.

Please also add a link to the `KernelLoadResult` structure so it is easier to follow the documentation.

src/loader/mod.rs
```diff
@@ -0,0 +1,580 @@
...
+            return Err(Error::InvalidElfMagicNumber);
+        }
+        if ehdr.e_ident[elf::EI_DATA as usize] != elf::ELFDATA2LSB as u8 {
+            return Err(Error::BigEndianElfOnLittle);
+        }
+        if ehdr.e_phentsize as usize != mem::size_of::<elf::Elf64_Phdr>() {
+            return Err(Error::InvalidProgramHeaderSize);
+        }
+        if (ehdr.e_phoff as usize) < mem::size_of::<elf::Elf64_Ehdr>() {
+            // If the program header is backwards, bail.
```
andreeaflorescu: What do you mean by backwards here?

src/loader/mod.rs
```diff
@@ -0,0 +1,580 @@
...
+        for phdr in &phdrs {
+            if phdr.p_type != elf::PT_LOAD || phdr.p_filesz == 0 {
+                continue;
+            }
+
+            kernel_image
+                .seek(SeekFrom::Start(phdr.p_offset))
+                .map_err(|_| Error::SeekKernelStart)?;
+
+            // vmm does not specify where the kernel should be loaded, just
```
andreeaflorescu: nit: If the VMM does not specify ... instead of just vmm does not specify ...

src/loader/mod.rs
```diff
@@ -0,0 +1,580 @@
...
+
+            guest_mem
+                .read_exact_from(mem_offset, kernel_image, phdr.p_filesz as usize)
+                .map_err(|_| Error::ReadKernelImage)?;
+
+            loader_result.kernel_end =
+                mem_offset.raw_value() as GuestUsize + phdr.p_memsz as GuestUsize;
+        }
+
+        loader_result.setup_header = None;
```
andreeaflorescu: Please explain why the setup_header is None in this case.

src/loader/mod.rs
```diff
@@ -0,0 +1,580 @@
...
+                    .ok_or(Error::InvalidProgramHeaderAddress)?,
+                None => GuestAddress(phdr.p_paddr as u64),
+            };
+
+            guest_mem
+                .read_exact_from(mem_offset, kernel_image, phdr.p_filesz as usize)
+                .map_err(|_| Error::ReadKernelImage)?;
+
+            loader_result.kernel_end =
+                mem_offset.raw_value() as GuestUsize + phdr.p_memsz as GuestUsize;
```
andreeaflorescu: Can this overflow? Should we use checked_add in that case?

src/loader/mod.rs
```diff
@@ -0,0 +1,580 @@
...
+/// * `kernel_load` - The actual `guest_mem` address where kernel image is loaded start.
+/// * `kernel_end` - The offset of `guest_mem` where kernel image load is loaded finish, return
+///                  in case of loading initrd adjacent to kernel image.
+/// * `setup_header` - The setup_header belongs to linux boot protocol, only for bzImage, vmm
+///                    will use it to setup setup_header.init_size, which is a must for bzImage
+///                    direct boot.
+#[derive(Debug, Default, Copy, Clone, PartialEq)]
+pub struct KernelLoaderResult {
+    pub kernel_load: GuestAddress,
+    pub kernel_end: GuestUsize,
```
andreeaflorescu: Why is `kernel_end` a usize while `kernel_load` is a u64?

src/loader/mod.rs
```diff
@@ -0,0 +1,580 @@
...
+    }
+}
+
+impl Display for Error {
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        write!(f, "Kernel Loader Error: {}", Error::description(self))
+    }
+}
+
+/// * `kernel_load` - The actual `guest_mem` address where kernel image is loaded start.
```
andreeaflorescu: The explanation of each field in the structure must be written before each field instead of at the beginning of the struct definition.

Can you also please rephrase the details to include the fact that kernel_load and kernel_end are addresses/offsets? Also, I wouldn't use `guest_mem` in the explanation because as an outside user of this crate you don't know what the `guest_mem` variable corresponds to.

```rust
#[derive(Debug, Default, Copy, Clone, PartialEq)]
pub struct KernelLoaderResult {
    /// Address in the guest memory where the kernel image is loaded.
    pub kernel_load: GuestAddress,
   /// Offset in guest memory corresponding to the end of the kernel image. TODO: add details about initrd. 
    pub kernel_end: GuestUsize,
    /// Optional header that specifies "TODO something something".
   /// This field is only used when loading bzimage and it is necessary because "something something"
    pub setup_header: Option<bootparam::setup_header>,
}
```

I don't know how bzimage works just yet, so i just wrote something something :))

src/loader/mod.rs
```diff
@@ -0,0 +1,580 @@
...
+/// * `kernel_end` - The offset of `guest_mem` where kernel image load is loaded finish, return
+///                  in case of loading initrd adjacent to kernel image.
```
andreeaflorescu: From this explanation here I understand the the kernel_end is only set when you also have an initrd. Is this assumption correct? In this case we should emphase this in the documentation of load_kernel because it is an important piece of information.

src/loader/mod.rs
```diff
@@ -0,0 +1,580 @@
...
+    /// * `lowest_kernel_start` - This is the start of the high memory, kernel should above it.
+    ///
+    /// # Returns
+    /// * KernelLoaderResult
+    fn load<F>(
+        guest_mem: &GuestMemoryMmap,
+        kernel_start: Option<GuestAddress>,
+        kernel_image: &mut F,
+        lowest_kernel_start: Option<GuestAddress>,
+    ) -> Result<(KernelLoaderResult)>
```
andreeaflorescu: Please remove the parentheses around `KernelLoaderResult`. They are not needed and on newer versions of rust fmt will fail. Same for the load implementation on Bzimage.

src/loader/mod.rs
bjzhjing: remove #[allow(dead_code)], is that what you mean?

src/loader/mod.rs
```diff
@@ -0,0 +1,580 @@
...
+///
+/// # Arguments
+///
+/// * `guest_mem` - A u8 slice that will be partially overwritten by the command line.
+/// * `guest_addr` - The address in `guest_mem` at which to load the command line.
+/// * `cmdline` - The kernel command line.
+pub fn load_cmdline(
+    guest_mem: &GuestMemoryMmap,
+    guest_addr: GuestAddress,
+    cmdline: &CStr,
```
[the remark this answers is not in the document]
bjzhjing: Good suggestion! I will do it once the current one merge.

src/loader/mod.rs
andreeaflorescu: I was asking why do you need to have `#[allow(dead_code)]`? I am thinking that you needed it because bootparam.rs looks like an autogenerated file and you might not use all the defines there and you are also not re-exporting them. So probably if you don't allow dead code the build will be full of warnings. I just wanted to check that's the case. It was more of a question than a change request.

bjzhjing: Yep, you're quite right! just because bootparam.rs is autogenerated and contains a bunch of unused code, add this to remove the build warnings.

src/loader/mod.rs
```diff
@@ -0,0 +1,580 @@
...
+#[derive(Debug, Default, Copy, Clone, PartialEq)]
+pub struct KernelLoaderResult {
+    pub kernel_load: GuestAddress,
+    pub kernel_end: GuestUsize,
+    pub setup_header: Option<bootparam::setup_header>,
+}
+
+pub trait KernelLoader {
+    fn load<F>(
+        guest_mem: &GuestMemoryMmap,
```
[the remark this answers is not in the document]
bjzhjing: Good question! VMM will pass mem to loader, so it's depends on VMM. If VMM use GuestMemoryMmap, but loader defined with GuestMemory, it will not work either. If VMM use GuestMemory, What GuestMemoryMmap is used for? I'm a bit confused. We need to have more discussions on this I think, event with other folks. I will just keep it as is at present.

src/loader/mod.rs
bjzhjing: Thanks! I prefer to highmem_start_address :-)

src/loader/mod.rs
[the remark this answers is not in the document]
andreeaflorescu: @bjzhjing are you joining the sync call today? I'll add this to the agenda.

src/loader/mod.rs
bjzhjing: means overlap. program header should follow after elf header,  it's invalid if not, that's what the above adjustment presents.

src/loader/mod.rs
bjzhjing: Add comment in code here.

src/loader/mod.rs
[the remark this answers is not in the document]
sameo: This should be generic indeed. Typically VMMs will pass mmap'ed memory to this call but there is no reason why this could not be generic and use `GuestMemory`.

src/loader/mod.rs
bjzhjing: sorry, what link do you mean?

src/loader/mod.rs
andreeaflorescu: Can you rephrase it so it's more clear what you want to check here? Program headers is backwards doesn't say much to me. Any reason you added a comment to this check, but didn't added comments to the other ones?

src/loader/mod.rs
andreeaflorescu: You can link the documentation of other structures in markdown.
Some examples here: https://github.com/rust-lang/rfcs/blob/master/text/1946-intra-rustdoc-links.md

src/loader/mod.rs
bjzhjing: Good point!

src/loader/mod.rs
bjzhjing: No, not exactly, other part, like setup_data could also be put there. I update the comment.

src/loader/mod.rs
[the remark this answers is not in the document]
bjzhjing: @sameo update with GuestMemory, but since cloud-hypervisor point to this crate, it requires changes there I think.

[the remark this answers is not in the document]
bjzhjing: @andreeaflorescu I update with GuestMemory, is it what you mean instead use traits (in slack)?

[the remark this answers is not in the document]
andreeaflorescu: Yes.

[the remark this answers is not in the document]
rbradford: @bjzhjing because of the `Cargo.lock` you won't break cloud-hypervisor. Go ahead and update this branch and then we can update in cloud-hypervisor after and bump the revision we use.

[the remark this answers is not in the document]
andreeaflorescu: @rbradford does this approach still work if you're doing push forces? The sha of the commit will be modified on push force. I noticed that you're using the same branch in cloud-hypervisor (master from @bjzhjing fork) as well. So unless you are using the cached repository to build cloud-hypervisor I think the build will fail.

src/loader/mod.rs
bjzhjing: There will be memory alignment calculation based on the returned kernel end for the next data struct to be put in, so just return GuestUsize directly.

src/loader/mod.rs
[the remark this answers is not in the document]
rbradford: @andreeaflorescu yes, it will fail upon a force push, that is a good point.

src/loader/mod.rs
bjzhjing: Nice!

.buildkite/hooks/post-checkout
```diff
@@ -0,0 +1,6 @@
+#!/bin/bash
+
```
andreeaflorescu: I am working on rust-vmm-ci. While writing the README.md I also wanted to give an example of a custom hook that can be placed in the repository. I took linux-loader as an example. You can use the code there to download the bzimage:
https://github.com/andreeaflorescu/rust-vmm-ci/tree/first_buildkite_pipeline#custom-repository-hooks

Some notes:
- We shouldn't rely on the name of the deb being downloaded. In my script I download the deb, extract it and copy the bzimage file to `linux-loader/src` with the name `bzimage`. This means that in the unit tests you will write bzimage instead of `./boot/vmlinuz-4.9.0-9-amd64`:

```rust
    fn make_bzImage() -> Vec<u8> {
        let mut v = Vec::new();
        v.extend_from_slice(include_bytes!("bzimage"));
        v
    }
```
- We should also take care of the clean-up; delete the downloaded files and only keep the bzimage you're actually using
- linux-4.9 is kinda ancient. should we use a newer image?

README.md
```diff
@@ -0,0 +1,105 @@
...
+Our Continuous Integration (CI) pipeline is implemented on top of
+[Buildkite](https://buildkite.com/).
+For the complete list of tests, check our
+[CI pipeline](https://buildkite.com/rust-vmm/vm-virtio-ci).
+
+Each individual test runs in a container. To reproduce a test locally, you can
+use the dev-container on both x86 and arm64.
+
+```bash
+docker run -it \
```
andreeaflorescu: Please use rustvmm/dev:v2 instead of fandree/rust-vmm-dev.

Cargo.toml
```diff
@@ -0,0 +1,16 @@
...
+elf = []
+bzImage = []
+
+
+[dependencies.vm-memory]
+git = "https://github.com/rust-vmm/vm-memory"
```
sboeuf: I think the dependency should be locked onto some specific revision, otherwise there is no way to tell when the crate will be broken by a change on vm-memory.

.buildkite/hooks/post-checkout
bjzhjing: @andreeaflorescu Good person! :-) yes, I will see to use a new version.

Cargo.toml
bjzhjing: I think it's sensible to point to rust-vmm/vm-memory branch. There are several PRs for vm-memory under review, once they are merged into master, we will fix it if the crate is broken by the dependency at that time. :-)

.buildkite/hooks/post-checkout
bjzhjing: @sameo  it seems in the container which docker started, it could not touch github, network connection failed, I guess it's caused by the lack of network proxy which we used in host. That's to say we currently can not do the test local.

tests/test_coverage.py
```diff
@@ -0,0 +1,108 @@
...
+    shutil.rmtree(kcov_build_dir, ignore_errors=True)
+
+    exclude_pattern = (
+        '${CARGO_HOME:-$HOME/.cargo/},'
+        'usr/lib/,'
+        'lib/'
+    )
+    exclude_region = "'mod tests {'"
+
+    kcov_cmd = "CARGO_TARGET_DIR={} cargo kcov --all " \
```
andreeaflorescu: you need to get the coverage for all features. cargo kcov --all --features=elf,bzImage

tests/test_coverage.py
```diff
@@ -0,0 +1,108 @@
...
+
+    # Remove kcov output and build directory to be sure we are always working
+    # on a clean environment.
+    shutil.rmtree(kcov_ouput_dir, ignore_errors=True)
+    shutil.rmtree(kcov_build_dir, ignore_errors=True)
+
+    exclude_pattern = (
+        '${CARGO_HOME:-$HOME/.cargo/},'
+        'usr/lib/,'
+        'lib/'
```
andreeaflorescu: you also need to exclude the autogenerated files: bootparams.rs and elf.rs

tests/test_coverage.py
```diff
@@ -0,0 +1,110 @@
...
+        '${CARGO_HOME:-$HOME/.cargo/},'
+        'usr/lib/,'
+        'lib/,'
+        'bootparam.rs,'
+        'elf.rs'
+    )
+    exclude_region = "'mod tests {'"
+
+    kcov_cmd = "CARGO_TARGET_DIR={} cargo kcov --all --features=elf,bzImage" \
+               "--output {} -- " \
```
rbradford: ^ Missing a space here

bjzhjing: yes, notice that, the newest one is on the way :-)

Cargo.toml
sboeuf: Well that's the point, to make sure this crate don't get broken by changes in vm-memory.
Right now, this crate works with the latest version of vm-memory, and we want this to continue to be true.
Once a new PR from vm-memory gets merged, if we want linux-loader to rely on the latest version, then this means we will need to submit a new PR to linux-loader to update the revision from vm-memory and potentially fix the code.

tests/test_coverage.py
bjzhjing: Done! all test cases have passed in buildkite for this crate.

Cargo.toml
bjzhjing: @sboeuf @rbradford @sameo 

I'm sincerely sorry, I just have chance to go through what is exactly done in the vendor PR sebastien create, linux-loader with 2b95f1e1958a2b6399b590f64344cab5b4a6d608 is pinned in cloud-hypervisor for use. But actually, it does not appear since I did git reset during recently changes. it broke the build for cloud-hypervisor, I'd like to update Cargo.toml with new existing commit, but seems lock file under .cargo need to  be removed firstly for new generation, which I could not touch directly. so I need your help to fix that. Really sorry for the inconvenience, I will pay attention on using git reset later.

Meanwhile, with the commit that remove dependency of GuestMemoryMmap, there is change required from cloud-hypervisor to pass GuestMemory to loader also. I could see how to change that if you need once commit get update in cloud-hypervisor.

andreeaflorescu: I don't think we need vendoring. Push forces are not allowed on the master branch of rust-vmm/vm-memory. What we do need is a fixed revision. We can either use a commit sha1 here or add tags to vm-memory and use tags in dependencies.

tests/test_coverage.py
```diff
@@ -0,0 +1,116 @@
...
+
+    # Remove kcov output and build directory to be sure we are always working
+    # on a clean environment.
+    shutil.rmtree(kcov_ouput_dir, ignore_errors=True)
+    shutil.rmtree(kcov_build_dir, ignore_errors=True)
+
+    # import pdb; pdb.set_trace()
+    crate_path = os.path.dirname(os.path.dirname(__file__))
+    parent_path = os.path.join(crate_path, "src", "loader")
+    bootparam_path = os.path.join(parent_path, "bootparam.rs")
```
andreeaflorescu: You don't need the full path. The exclude path is a regex so it will exclude the `bootparams.rs` file even if you just write `bootparam.rs`.

tests/coverage
```diff
@@ -0,0 +1 @@
+71.3
```
andreeaflorescu: I find this coverage a bit weird. I run it locally with the coverage test from rust-vmm-ci and it give me a coverage of 82.9. Maybe you're not properly excluding files?

sameo:
> Still missing from this PR:
> 
> * documentation on public interface is not enforced with `#![deny(missing_docs)]` in `src/lib.rs`
> * properly mark the code as x86_64 specific and add pipeline steps for aarch64 as well.
> * musl tests in pipeline (we can open an issue about this one and add them in the next PR)
> 
This should all be addressed now. We have 2 buildkite pipeline, one based on the pending rust-vmm-ci one and a linux-loader specific one.

sameo:
And it's all green.

.buildkite/pipeline.yml
```diff
@@ -0,0 +1,26 @@
+# Copyright 2019 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+# SPDX-License-Identifier: Apache-2.0
+steps:
+  - label: "build-gnu-x86"
```
andreeaflorescu: I think you still need to keep these steps:
```yml
 - label: "build-gnu-x86-elf"
    commands:
     - cargo build --release --features elf
    retry:
      automatic: false
    agents:
      platform: x86_64.metal
    plugins:
      - docker#v3.0.1:
          image: "fandree/rust-vmm-dev"
          always-pull: true

  - label: "unittests-gnu-x86-elf"
    commands:
     - cargo test --features elf
    retry:
      automatic: false
    agents:
      platform: x86_64.metal
    plugins:
      - docker#v3.0.1:
          privileged: true
          image: "fandree/rust-vmm-dev"

  - label: "build-gnu-x86-bzImage"
    commands:
     - cargo build --release --features bzImage
    retry:
      automatic: false
    agents:
      platform: x86_64.metal
    plugins:
      - docker#v3.0.1:
          image: "fandree/rust-vmm-dev"
          always-pull: true

  - label: "unittests-gnu-x86-bzImage"
    commands:
     - cargo test --features bzImage
    retry:
      automatic: false
    agents:
      platform: x86_64.metal
    plugins:
      - docker#v3.0.1:
          privileged: true
          image: "fandree/rust-vmm-dev"
```
because you want to check that the build passes with individual features as well. Actually, since elf is the default feature now, you can skip the elf build step and only add the bzImage one.

sameo: Done.
I did not realize we switched to using `ELF` as the default feature.

.buildkite/pipeline.yml
```diff
@@ -0,0 +1,52 @@
...
+    retry:
+      automatic: false
+    agents:
+      platform: x86_64.metal
+    plugins:
+      - docker#v3.0.1:
+          image: "rustvmm/dev:v2"
+          always-pull: true
+
+  - label: "unittests-gnu-x86-bzImage"
```
andreeaflorescu: This one is not needed because we are unit testing all the features with the default pipeline from rust-vmm-ci. Same for musl unittests.

Cargo.toml
andreeaflorescu: @sameo @bjzhjing do you want to use a fixed revision in this PR or do you want to fix it in a future PR?

rbradford:
I think the bzImage feature should be renamed "bzimage" as the elf feature is not ELF. And as AFAICT most other crates use entirely lowercase features.

Cargo.toml
```diff
@@ -0,0 +1,15 @@
+[package]
+name = "linux-loader"
+version = "0.1.0"
+authors = ["cathy <cathy.zhang@intel.com>"]
```
rbradford: I think this needs to be your complete name.

andreeaflorescu:
@rbradford yap, I agree. We should do the same for function names as well (i.e. make_bzImage, load_bzImage) and then we can also get rid of all the `allow(non_snakecase)`

Cargo.toml
```diff
@@ -0,0 +1,15 @@
+[package]
+name = "linux-loader"
+version = "0.1.0"
```
rbradford: @andreeaflorescu is this correct starting version number?

DESIGN.md
```diff
@@ -0,0 +1,64 @@
...
+It follows ELF standard which is specified in elf.rs. The entry header and program headers will be inerpreted, and PT_LOAD segments will be loaded into guest memory.
+
+### Where kernel is loaded 
+
+There are two ways on deciding where the program segments will be loaded.
+
+- One way is to provide an option and allow vmm to specify where to load the image, considering its memory layout.
+- The other way is to load image into phdr.p_paddr by default.
+
+## Vmm side
```
rbradford: VMM is an acronym,

DESIGN.md
```diff
@@ -0,0 +1,64 @@
...
+
+### Construct zero page
+
+According to the 64-bit boot protocol, the boot parameters (traditionally known as "zero page") should be setup, including setup_header, e820 table and other stuff. However, ELF has no setup_header, nothing returned from ELF loader could be used to fill boot parameters, vmm is totally responsible for the construction. 
+
+### Configure vcpu
```
rbradford: Usually written vCPU.

rbradford:
> @rbradford yap, I agree. We should do the same for function names as well (i.e. make_bzImage, load_bzImage) and then we can also get rid of all the `allow(non_snakecase)`

:+1:

Cargo.toml
andreeaflorescu: I would have it as 0.0.1 and when we publish it to crates.io I would switch to 0.1.0. But it doesn't matter that much if we don't update the version on following PRs.

DESIGN.md
sameo: Fixed.

Cargo.toml
sameo: Fixed

sameo:
> @rbradford yap, I agree. We should do the same for function names as well (i.e. make_bzImage, load_bzImage) and then we can also get rid of all the `allow(non_snakecase)`

Done.

sameo:
> I think the bzImage feature should be renamed "bzimage" as the elf feature is not ELF. And as AFAICT most other crates use entirely lowercase features.

Done.

bjzhjing:
@rbradford just send the PR against cloud-hypervisor with adding bzimage support.

src/loader/mod.rs
```diff
@@ -0,0 +1,650 @@
...
+    fn load<F, M: GuestMemory>(
+        guest_mem: &M,
+        kernel_start: Option<GuestAddress>,
+        kernel_image: &mut F,
+        highmem_start_address: Option<GuestAddress>,
+    ) -> Result<KernelLoaderResult>
+    where
+        F: Read + Seek,
+    {
+        let mut kernel_size = kernel_image
```
rbradford: I think you're missing this step for the bzImage loading process

From https://www.kernel.org/doc/Documentation/x86/boot.txt

```In 32-bit boot protocol, the first step in loading a Linux kernel
should be to setup the boot parameters (struct boot_params,
traditionally known as "zero page"). The memory for struct boot_params
should be allocated and initialized to all zero. Then the setup header
from offset 0x01f1 of kernel image on should be loaded into struct
boot_params and examined```

The load function function prototype should probably be modified to take a `boot_params`

rbradford: I missed that this header is being returned which should be sufficient to satisfy this requirement.

PR #6, opened 2019-08-14
loader: Regenerate bootparams.rs

rbradford:
Run rust-bindgen against the latest stable version of bootparams.h
(v5.2 tag - 0ecfebd2b52404ae0c54a878c872bb93363ada36)

Generated with:

bindgen bootparam.h -o bootparam.rs --with-derive-default
--with-derive-partialeq --no-doc-comments

Bindgen version: 0.51.0


rbradford:
@andreeaflorescu here is an alternative you might prefer.

src/loader/bootparam.rs
```diff
@@ -3861,10 +3868,47 @@ impl Default for boot_params {
...
-pub const X86_SUBARCH_CE4100: _bindgen_ty_1 = 4;
-pub const X86_NR_SUBARCHS: _bindgen_ty_1 = 5;
-pub type _bindgen_ty_1 = ::std::os::raw::c_uint;
+pub const x86_hardware_subarch_X86_SUBARCH_PC: x86_hardware_subarch = 0;
+pub const x86_hardware_subarch_X86_SUBARCH_LGUEST: x86_hardware_subarch = 1;
+pub const x86_hardware_subarch_X86_SUBARCH_XEN: x86_hardware_subarch = 2;
+pub const x86_hardware_subarch_X86_SUBARCH_INTEL_MID: x86_hardware_subarch = 3;
+pub const x86_hardware_subarch_X86_SUBARCH_CE4100: x86_hardware_subarch = 4;
+pub const x86_hardware_subarch_X86_NR_SUBARCHS: x86_hardware_subarch = 5;
+#[doc = " enum x86_hardware_subarch - x86 hardware subarchitecture"]
```
andreeaflorescu: These doc entries are a bit weird. Any way we can disable them? I don't think they are necessarily needed and they don't look particularly readable either.

rbradford: I was able to get rid of them with a bindgen parameter.

rbradford:
> Please keep the 50/70 rule when you write the commit message. Summary: 50 chars and each line of the description should have maximum 70 chars.
> 
> What helped us a lot in the past in Firecracker was to also specify the bindgen version used for generating the file. We can specify the version in the commit message as well as it is easy to find that information with a `git log`.

Done.

PR #9, opened 2019-11-26
Support raw image (vmlinux) loading for Aarch64

michael2012z:
Add support for Aarch64 to load vmlinux in PE format.

Reference to this Firecracker [commit](https://github.com/firecracker-microvm/firecracker/commit/ae519f57ea39ccf18f4cc35488a583102fb5c191).

Note: In unit test, a binary file ("src/loader/test_arm64_pe.bin") is loaded. (For x86 test, it is "test_elf.bin") But as per our company's policy, I can't upload any binary file to a public repository. Please any maintainer help download the file from [here](https://github.com/firecracker-microvm/firecracker/blob/master/kernel/src/loader/test_pe.bin) and upload. Thanks.

Signed-off-by: Michael Zhao <michael.zhao@arm.com>


michael2012z:
Hi, @andreeaflorescu @sameo , no reviewer was assigned automatically. Could you help review this PR?

src/loader/mod.rs
```diff
@@ -12,19 +12,20 @@
...
 
 extern crate vm_memory;
 
 use std::error::{self, Error as KernelLoaderError};
 use std::ffi::CStr;
 use std::fmt::{self, Display};
-#[cfg(any(feature = "elf", feature = "bzimage"))]
-#[cfg(any(target_arch = "x86", target_arch = "x86_64"))]
+#[cfg(any(feature = "elf", feature = "bzimage", feature = "arm64_pe"))]
+#[cfg(any(target_arch = "x86", target_arch = "x86_64", target_arch = "aarch64"))]
```
aghecenco: These 2 lines cover all the features and all the supported architectures, you could remove them to keep the code simple

src/loader/mod.rs
```diff
@@ -12,19 +12,20 @@
...
 use std::io::SeekFrom;
 use std::io::{Read, Seek};
-#[cfg(feature = "elf")]
-#[cfg(any(target_arch = "x86", target_arch = "x86_64"))]
+#[cfg(any(feature = "elf", feature = "bzimage", feature = "arm64_pe"))]
+#[cfg(any(target_arch = "x86", target_arch = "x86_64", target_arch = "aarch64"))]
```
aghecenco: Same observation as above

Cargo.toml
```diff
@@ -9,6 +9,7 @@ license = "Apache-2.0 AND BSD-3-Clause"
 default = ["elf"]
 elf = []
 bzimage = []
+arm64_pe = []
```
aghecenco: Please add a TODO & an issue so we can add target-dependent features when https://github.com/rust-lang/cargo/issues/1197 makes it into `stable`. I'd like to see the `arm64_pe` only available on `aarch64` (and maybe just called `pe`).

src/loader/mod.rs
```diff
@@ -81,6 +80,10 @@ pub enum Error {
     ReadBzImageHeader,
     /// Unable to read bzImage compressed image.
     ReadBzImageCompressedKernel,
+    /// Unable to read image magic number.
+    ReadImageMagicNumber,
```
aghecenco: Would it be useful to extend the error with the bad magic number?

src/loader/mod.rs
```diff
@@ -81,6 +80,10 @@ pub enum Error {
...
+    /// Unable to read entry address.
+    ReadEntryAddress,
```
aghecenco: Same obs. as above: I'm thinking of debugging possibilities for a kernel loading issue.

src/loader/mod.rs
```diff
@@ -378,6 +395,117 @@ impl KernelLoader for BzImage {
     }
 }
 
+#[cfg(feature = "arm64_pe")]
```
aghecenco: nit: to shorten this, you could chain the `cfg`s: `#[cfg(all(feature = "arm64_pe", target_arch = "aarch64"))]

aghecenco: (it would be even better if target-specific features were available :( )
BTW, nit-ish observation, I'd like to keep the target consistent with the feature name (i.e. call the feature `aarch64_pe` instead, or just plain `pe`), WDYT?

src/loader/mod.rs
```diff
@@ -378,6 +395,117 @@ impl KernelLoader for BzImage {
...
+    /// u64 res3  = 0;            /* reserved */
+    /// u64 res4  = 0;            /* reserved */
+    /// u32 magic = 0x644d5241;   /* Magic number, little endian, "ARM\x64" */
+    /// u32 res5;                 /* reserved (used for PE COFF offset) */
+    /// ====================================
+    ///
+    /// # Arguments
+    ///
+    /// * `guest_mem` - The guest memory region the kernel is written to.
+    /// * `kernel_start` - The offset into 'guest_mem' at which to load the kernel.
```
aghecenco:
```suggestion
    /// * `kernel_start` - The offset into `guest_mem` at which to load the kernel.
```

src/loader/mod.rs
```diff
@@ -378,6 +395,117 @@ impl KernelLoader for BzImage {
...
+    /// * `kernel_image` - Input vmlinux image.
+    /// * `highmem_start_address` - This is the start of the high memory, kernel should above it.
```
aghecenco:
```suggestion
    /// * `highmem_start_address` - Start of the high memory.
```

src/loader/mod.rs
```diff
@@ -378,6 +395,117 @@ impl KernelLoader for BzImage {
...
+    ///
+    /// # Returns
```
aghecenco: I think we could eliminate the `# Returns` paragraphs to be consistent with other rust-vmm crates.

src/loader/mod.rs
```diff
@@ -378,6 +395,117 @@ impl KernelLoader for BzImage {
...
+    fn load<F, M: GuestMemory>(
+        guest_mem: &M,
+        kernel_start: Option<GuestAddress>,
+        kernel_image: &mut F,
+        _highmem_start_address: Option<GuestAddress>,
+    ) -> Result<KernelLoaderResult>
+    where
+        F: Read + Seek,
+    {
+        const AARCH64_KERNEL_LOAD_ADDR: usize = 0x80000;
```
aghecenco: These `const`s might be better placed outside the function.

aghecenco:
@michael2012z you can cherry pick https://github.com/aghecenco/linux-loader/commit/d4b972073913c10bfe2c02865778404b6d7e89c7 for the kernel binary. The unit tests pass with this commit, however there are a bunch of build warnings on `aarch64` that would need cleaning up too. Would you like to take a look at those as well?

michael2012z:
@aghecenco Thank you very much for review and adding the binary. I will come back soon to update this long-waiting PR.

michael2012z:
Hi, @aghecenco , the file `test_pe.bin` turns a HTML text file: https://github.com/aghecenco/linux-loader/blob/aarch64pe/src/loader/test_pe.bin

Could you help again?

aghecenco:
> Hi, @aghecenco , the file `test_pe.bin` turns a HTML text file: https://github.com/aghecenco/linux-loader/blob/aarch64pe/src/loader/test_pe.bin
> 
> Could you help again?

Oops! Cherry-pick  https://github.com/aghecenco/linux-loader/commit/2175ff88c010da6836968b7968751ff8a139d352 instead. I updated it.

Cargo.toml
michael2012z: Before target-dependent feature come to stable, is there a workaround in our CI? For arm build and test, now we need to enable `arm64_pe` but disable `elf`. 
In local test, I do with command `cargo test --features arm64_pe --no-default-features`.

src/loader/mod.rs
michael2012z: Good idea to align all the names to `aarch64*`.

src/loader/mod.rs
michael2012z: I made them trait-level `const`'s.

dianpopa:
@aghecenco @michael2012z  i strongly encourage you to implement the kernel loading protocols in separate sources (even folders) per architecture.
Right now the hierarchy of the crate looks like this:
root
        -> mod.rs (contains kernel loading protocol for **all** supported archs including unit tests)
        -> elf.rs (**x86_64** specific)
        -> struct_util.rs (**x86_64 specific**)
        -> bootparam.rs (**x86_64 specific**)
        ->  test_elf.bin (**x86_64 specific**)

root
      -> x86_64 -> mod.rs, elf.rs, struct_util.rs, bootparam.rs
       -> aarhc64 -> mod.rs, test_arm64_pe.bin

I think the second proposal would reflect a much cleaner hierarchy, would make the code much easier to read since it gets rid of lots of cfg syntax and it is also encourages extensibility.

src/loader/mod.rs
```diff
@@ -55,8 +50,8 @@ pub enum Error {
     CommandLineCopy,
     /// Command line overflowed guest memory.
     CommandLineOverflow,
-    /// Invalid ELF magic number
-    InvalidElfMagicNumber,
+    /// Invalid magic number
+    InvalidMagicNumber,
```
michael2012z: I renamed it to from `InvalidElfMagicNumber` to `InvalidMagicNumber`, a more generic one. Because an error message is needed when parsing ARM kernel, but defining a new `InvalidPeMagicNumber` looks too verbose.

michael2012z:
> @aghecenco @michael2012z i strongly encourage you to implement the kernel loading protocols in separate sources (even folders) per architecture.

This echos an earlier comment here by @aghecenco : https://github.com/rust-vmm/linux-loader/pull/9#pullrequestreview-370460221

That sounds a good idea. Now ARM code has very little in common with X86 staff.

src/loader/mod.rs
michael2012z: Removed both. But I added a new one `InvalidPeImage` for common parsing error.

Cargo.toml
michael2012z: Will create an issue to track it after merging.

michael2012z:
@dianpopa , @aghecenco , Thanks for review the code. I handled the comments about code that have been placed so far. And errors and warnings were all cleared.

Next I am going to refactor the repo to divide source code files by architecture. I tried that a bit, looks like some effort is needed to finish. Considering now there are 2 PR's doing the same job (https://github.com/rust-vmm/linux-loader/pull/16 and this), I think it's better to merge one of them at first to avoid investing more effort in a PR but it is finally given up.

aghecenco:
@michael2012z I compared the changes with #16 and to me the image loading part looks identical functionality-wise, but it's slightly compact in the newer PR and I expect it to be easier to rebase #26 on top of it. There's also the added device tree blob loader. Would you mind closing this in favor of #16 and reviewing it, as you're clearly more familiar with the topic?

michael2012z:
Hi, @aghecenco , no problem, I will close it. Sure, I will continue working in this crate. 
And thank you very much for reviewing this PR.

PR #11, opened 2020-02-20
Add support for PVH boot protocol

### LINKED ISSUE #3: Support PVH boot protocol
The PVH boot protocol allows loading an uncompressed kernel instead of a bzImage. This is faster as it removes decompression from the total startup time.

aljimenezb:
Adding support for parsing the PVH entry point address if encoded in the kernel ELF binary, and returning it as an additional field in KernelLoaderResult. This allows the VMM to choose which boot protocol to use.

These changes have been tested alongside with patches for Intel Cloud Hypervisor that successfully boot a guest using the PVH entry point. 

The PR is marked as WIP since there are still a few open questions:
- Some definitions from Xen's xen/include/public/arch-x86/hvm/start_info.h (this header also exists in Linux), and comments were manually copied into a new file start_info.rs. When using bindgen to generate start_info.rs from the C header, the output file is not very legible so I opted for the manual approach to keep it readable. Is this acceptable or should bindgen be used?

- I took the function align_up() from the x86_64 crate (https://docs.rs/x86_64/0.9.2/x86_64/index.html) and slightly modified it. The goal was to avoid adding a new dependency to an external crate. Is this appropriate?

- Copyright messages: I'm unsure of the proper way, so I just pasted the copyright notice at the top of every file I touched, regardless of the length/importance of the change. What are the guidelines on how to do this correctly?

Closes #3


rbradford:
@bonzini Can you give this a quick look through?

bonzini:
Looks good to me. The main issue is that it does not cover the task of preparing the PVH parameters struct, I'll open an issues about that.

src/loader/elf.rs
```diff
@@ -86,6 +98,21 @@ impl Clone for elf64_phdr {
...
+    pub n_namesz: Elf64_Word, // Name size
+    pub n_descsz: Elf64_Word, // Content size
+    pub n_type: Elf64_Word,   // Content type
+}
+
+impl Clone for elf64_nhdr {
+    fn clone(&self) -> Self {
+        *self
+    }
+}
```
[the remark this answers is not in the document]
aljimenezb: Will do. I was trying to follow the example from the lines above which explicitly define clone for the other headers. e.g.
`impl Clone for elf64_phdr {`

but that was generated by bindgen, so I'll do the "human" thing as you suggested.

aljimenezb:
> Looks good to me. The main issue is that it does not cover the task of preparing the PVH parameters struct, I'll open an issues about that.

Right, I looked to see if there was some common place to add this code (and also setting up initial vCPU registers), since these areas are very similar in Firecracker and Intel Cloud Hypervisor (likely inherited from crosvm). I thought the vmm-vcpu crate would be a good place for this but found it empty. Looking at the issue you opened now..

andreeaflorescu:
We also need to add tests for the newly added feature. For some examples, you can check the existing pipeline: https://github.com/rust-vmm/linux-loader/blob/master/.buildkite/pipeline.yml

It's not a hard block, but I would prefer the PR that uses the rust-vmm-ci pipeline to be merged first because it also updates the container (which updates the rust version & adds tests for aarch64): https://github.com/rust-vmm/linux-loader/pull/14

aljimenezb:
@andreeaflorescu I amended the last commit to use an ELF binary with an additional note header that is unrelated to PVH. The goal is to increase coverage by exercising the code path that ignores such notes. Still, the coverage test fails because coverage drops by 0.20% for x86 and 1% for arm.

I believe the minor drops in coverage are due to the start_info.rs definitions which do not have layout tests, but I am not sure how the tests work exactly. The previous test left some files in the workspace where I could examine the report and see the coverage, but after updating to the rust-vmm-ci pipeline I can't find those anymore. Any advice on what I am doing wrong/how to move forward?

aljimenezb:
Added layout tests (autogenerated by bindgen) for the hvm_start_info and hvm_memmap_table_entry structs. Also changed the name from hvm_**mmap**\_table_entry to hvm_**memmap**\_table_entry to keep the same name as in the Xen header file (https://xenbits.xenproject.org/docs/unstable/hypercall/x86_64/include,public,arch-x86,hvm,start_info.h.html)

To investigate why the coverage check fails with a drop of 0.20% for x86 and 1% for arm, I modified test_coverage.py in my local repo so the kcov_output directory is not cleaned up after the test runs. I can see that in the x86 case, there are only 3 lines I added that are not covered.

src/loader/mod.rs
```diff
@@ -280,6 +301,73 @@ impl KernelLoader for Elf {
...
+            + align_up(u64::from(nhdr.n_descsz), n_align);
+
+        kernel_image
+            .seek(SeekFrom::Start(phdr.p_offset + read_size as u64))
+            .map_err(|_| Error::SeekNoteHeader)?;
+    }
+
+    if read_size >= phdr.p_filesz as usize {
+        return Ok(None); // PVH ELF note not found, nothing else to do.
+    }
```
aljimenezb: Line 344 is missing coverage, it is only executed if there are note headers, but none of them contain a PVH note.

src/loader/mod.rs
```diff
@@ -280,6 +301,73 @@ impl KernelLoader for Elf {
...
+        ))
+        .map_err(|_| Error::SeekNoteHeader)?;
+
+    // We support 64bit kernels only and the PVH entry point is a 32bit address
+    // encoded in a 64 bit field, so we'll grab all 8 bytes.
+    let mut pvh_addr_bytes = [0; 8usize];
+
+    if nhdr.n_descsz as usize != pvh_addr_bytes.len() {
+        return Err(Error::ReadNoteHeader);
+    }
```
aljimenezb: Line 360 is missing coverage. This is a check to prevent a malformed PVH note header from being returned as a valid address.

aljimenezb:
I found an ARM box and ran the coverage tests with the instrumented test_coverage.py so that I could look at the reports afterwards.

Before applying changes in this PR:
Instrumented lines: 1039
Executed lines: 737
Code covered: 70.9%

Filename | Coverage % | Covered lines | Uncovered lines | Executable lines
-- | -- | -- | -- | --
/linux-loader/src/loader/elf.rs | 0.0% | 0 | 30 | 30
/linux-loader/src/loader/mod.rs | 53.1% | 43 | 38 | 81
/linux-loader/src/loader/bootparam.rs | 74.1% | 628 | 219 | 847
/linux-loader/src/cmdline/mod.rs | 81.5% | 66 | 15 | 81

=======================================================
After applying changes:
Instrumented lines: 1062
Executed lines: 743
Code covered: 70.0%

Filename | Coverage % | Covered lines | Uncovered lines | Executable lines
-- | -- | -- | -- | --
/linux-loader/src/loader/start_info.rs | 0.0% | 0 | 16 | 16
/linux-loader/src/loader/elf.rs | 0.0% | 0 | 34 | 34
/linux-loader/src/loader/mod.rs | 57.1% | 48 | 36 | 84
/linux-loader/src/loader/bootparam.rs | 74.1% | 628 | 219 | 847
/linux-loader/src/cmdline/mod.rs | 82.7% | 67 | 14 | 81

(Ignoring the spurious change in /linux-loader/src/cmdline/mod.rs, which is caused by an empty line which is now reported as covered)

The results above show that the drop in coverage for the aarch64 case is caused primarily by the new data structures in start_info.rs and elf.rs which are not utilized in the aarch64 code.
Given that as it is now, PVH is purely an x86 specification, we can restrict the inclusion of the start_info structures to the x86 builds.

andreeaflorescu:
> Given that as it is now, PVH is purely an x86 specification, we can restrict the inclusion of the start_info structures to the x86 builds.

Sounds good, let's compile out what is not used on aarch64. If the coverage still fails for weird reasons (it does that from time to time), you can also decrease the coverage.

aljimenezb:
> Sounds good, let's compile out what is not used on aarch64. If the coverage still fails for weird reasons (it does that from time to time), you can also decrease the coverage.

Done. I tried my best to avoid decreasing the coverage, and luckily the numbers worked out after removing start_info.rs from the aarch64 case, and adding more unit tests on x86. I actually had to increase the coverage value for x86 on my last commit. Please let me know if there is anything else that is needed before merging.

aghecenco:
I would very much like to see https://github.com/rust-vmm/linux-loader/pull/24 merged before this one, so that we can work on code that's clearly structured per architecture and kernel format (this code would then go to `x86/elf` afaict) (and we shouldn't have confusing coverage per-arch anymore). However, if it's functionally merge-able, I won't block it, I'll rebase #24 over instead.

aljimenezb:
> I would very much like to see #24 merged before this one, so that we can work on code that's clearly structured per architecture and kernel format (this code would then go to `x86/elf` afaict) (and we shouldn't have confusing coverage per-arch anymore). However, if it's functionally merge-able, I won't block it, I'll rebase #24 over instead.

Unless anyone else has objections, I believe this PR is ready to merge, but I'm ok with whatever you decide is best. As you mentioned, I think the PVH code probably fits best in `x86/elf` given that is a special case of ELF boot. Not sure that a dedicated `x86_64/pvh` directory is needed...

aghecenco:
>  I think the PVH code probably fits best in `x86/elf` given that is a special case of ELF boot. Not sure that a dedicated `x86_64/pvh` directory is needed...

After actually *looking at the code*, I realize that it isn't. I will update the description.

src/loader/start_info.rs
```diff
@@ -0,0 +1,298 @@
...
+ *    | reserved       |
+ * 24 +----------------+
+ *
+ * The address and sizes are always a 64bit little endian unsigned integer.
+ */
+
+// Rust definitions needed to enable PVH boot protocol
+#[repr(C)]
+#[derive(Debug, Copy, Clone, Default)]
+pub struct hvm_start_info {
```
andreeaflorescu: **disclaimer** I don't really know how this works.

I noticed that the structures in this file aren't actually used. Is that intended? How are you suppose to use them? I noticed that we do export them, so I was wondering if we can have some sort of example with them.

andreeaflorescu:
@bonzini can you take a look at this PR?

src/loader/start_info.rs
aljimenezb: The structures are meant to be exported and used by the VMM implementation when writing the necessary info (cmdline address, ACPI RSDP, memory maps, etc) to guest memory to implement the specific boot protocol. A current example is `struct boot_params`, which is also defined in linux-loader and then imported by the VMMs. If you were to try and stablish a correspondence between the two, then:
hvm_start_info <==> boot_params
hvm_memmap_table_entry <==> boot_e820_entry

As Paolo mentioned before, PVH uses smaller set of boot parameters than the Linux boot protocol; you can see by the definitions that hvm_start_info is much more concise than boot_params.

Currently, VMMs like Intel Cloud Hypervisor and Firecracker use the Linux boot protocol and `struct boot_params`.  The goal of this change is to allow them to use the PVH protocol as well, since it is a standard Linux interface and boots as fast (or faster) than the direct ELF boot method currently in use. I have an open [PR](https://github.com/cloud-hypervisor/cloud-hypervisor/pull/801) on Intel Cloud Hypervisor and will create one for Firecracker soon.

Right now, for both direct ELF boot and PVH, the VMM is responsible for putting the data structures in guest memory and initializing the vCPU state as required by the specific protocol. Paolo opened https://github.com/rust-vmm/linux-loader/issues/15 to add functionality that takes care of those tasks as well on this crate. IIUC, that would provide the sort of use case example that is missing for both boot_params and hvm_start_info and friends... And of course I can also add comments on start_info.rs if you think that is better.

andreeaflorescu: Thanks for the really nice explanation. I think it is worth specifying this somewhere so that noobs like myself can have it easier.

src/loader/elf.rs
```diff
@@ -22,6 +24,16 @@ pub const ELFMAG0: ::std::os::raw::c_uint = 127;
 pub const ELFDATA2LSB: ::std::os::raw::c_uint = 1;
 pub const PT_LOAD: ::std::os::raw::c_uint = 1;
 
+// ELF Note p_type used to parse PVH entry point address
```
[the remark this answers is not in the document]
aghecenco: (Note that if you end up restructuring like this, `bootparam.rs` needs to suffer the same fate.)

src/loader/start_info.rs
```diff
@@ -0,0 +1,298 @@
...
+    pub type_: u32,
+    pub reserved: u32,
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[test]
+    fn bindgen_test_layout_hvm_start_info() {
```
aghecenco: Is this file generated with bindgen? If so, please add a comment like in the other generated files (elf.rs, bootparam.rs). If partially, please add instructions on how to replicate it.

src/loader/mod.rs
```diff
@@ -280,6 +303,72 @@ impl KernelLoader for Elf {
...
+
+    kernel_image
+        .read_exact(&mut pvh_addr_bytes)
+        .map_err(|_| Error::ReadNoteHeader)?;
+
+    Ok(Some(GuestAddress(<u64>::from_le_bytes(pvh_addr_bytes))))
```
aghecenco:
```suggestion
    Ok(Some(GuestAddress(u64::from_le_bytes(pvh_addr_bytes))))
```

src/loader/mod.rs
```diff
@@ -280,6 +303,72 @@ impl KernelLoader for Elf {
...
+        .seek(SeekFrom::Current(
+            align_up(u64::from(nhdr.n_namesz), n_align) as i64,
+        ))
+        .map_err(|_| Error::SeekNoteHeader)?;
+
+    // We support 64bit kernels only and the PVH entry point is a 32bit address
+    // encoded in a 64 bit field, so we'll grab all 8 bytes.
+    let mut pvh_addr_bytes = [0; 8usize];
+
+    if nhdr.n_descsz as usize != pvh_addr_bytes.len() {
```
aghecenco: This can be checked above, before declaring the byte buffer, if you compare `nhdr.n_descsz` with 8.

src/loader/mod.rs
```diff
@@ -409,6 +498,22 @@ pub fn load_cmdline<M: GuestMemory>(
     Ok(())
 }
 
+/// Align address upwards. Taken from x86_64 crate.
```
aghecenco:
```suggestion
/// Align address upwards. Taken from x86_64 crate: https://docs.rs/x86_64/latest/x86_64/fn.align_up.html
```

src/loader/elf.rs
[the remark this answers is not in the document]
aljimenezb: > This file is autogenerated by `bindgen`. I'd prefer avoiding to edit it directly. How about renaming the original `elf.rs` to `elf_gen.rs` or in a different `elf_gen` module altogether, and adding the PVH-specific definitions in a standalone, non-generated file?

My argument for modifying this file and not creating new ones is because `elf.rs` was autogenerated by `bindgen` but it seems to have been manually (and heavily) edited before arriving at this current version. 
Checking out upstream linux commit `806276b7f07a39a1cc3f38bb1ef5c573d4594a38` and looking at `linux include/uapi/linux/elf.h` shows that PT_NOTE and Note header definitions were already present in the upstream source when `elf.rs` was generated by `bindgen`. These definitions and many others were probably pruned to keep the file small since they were not needed at the time, but from a logical standpoint they should live in this file.

If you agree with my argument above, then perhaps the better solution is to run `bindgen` again on `include/uapi/linux/elf.h`, but I think that is out of the scope for this specific PR. If you don't agree, please let me know and I'll create a new `elfnote.rs` to contain the Note definitions I am adding.

Just to clarify, I did include the `pub const XEN_ELFNOTE_PHYS32_ENTRY: u32 = 18;` definition here that comes from a different header (`include/xen/interface/elfnote.h`) since I couldn't think of a better place for it and thought that it fit in the same scope as the Note definitions.

src/loader/start_info.rs
aljimenezb: Partially. I used bindgen initially but the output is not very legible, so I opted for manually copying the comments and the definitions I needed into start_info.rs. The tests I added later when trying to figure out the coverage failure.

I'll add instructions.

src/loader/mod.rs
aljimenezb: Will fix it. I'm actually changing those lines a bit, to use more generic code that is not restricted to just 64-bit kernels. I''ll post changes soon.

src/loader/mod.rs
aljimenezb: Fixed in last push.

src/loader/mod.rs
aljimenezb: Changed that code slightly in the last push, I'd appreciate it if you could review the new version

src/loader/mod.rs
aljimenezb: Resolved.

src/loader/start_info.rs
aljimenezb: Generated `start_info.rs` again using `bindgen`, grabbing the `start_info.h` header from Xen tagged release 4.13.0. Made these changes to the generated start_info.rs:

- Kept the original Citrix copyright stanza.
- Listed bindgen command and upstream commit used to generate the file as requested by @aghecenco.
- Added comment detailing purpose of the definitions as requested by @andreeaflorescu.
- Kept comments from C header with structure layout and explanations.
- Included the hvm_modlist_entry definition which is needed for supporting initrd in PVH.
- Added Default to the list of traits that are automatically derived for each of the definitions.
- Grouped definitions in the default module and moved all the tests to test module.
- Edited the file for clarity, trimming unnecessary definitions.

Please see newest version and let me know if anything is missing.

src/loader/mod.rs
```diff
@@ -280,6 +306,75 @@ impl KernelLoader for Elf {
...
+    let mut pvh_addr_bytes = [0; mem::size_of::<u32>()];
+
+    // Read 32-bit address stored in the PVH note descriptor field.
+    kernel_image
+        .read_exact(&mut pvh_addr_bytes)
+        .map_err(|_| Error::ReadNoteHeader)?;
+
+    Ok(Some(GuestAddress(
+        u32::from_le_bytes(pvh_addr_bytes).into(),
+    )))
```
aljimenezb: I changed this section slightly since the previous revision. The previous code works correctly for 64-bit Linux ELF header which encodes the address in an 8 byte field, but it did not worked with ELF binaries that could encode the PVH address using a 4 byte field.
The PVH specification technically does not define the size of the note header descriptor field, only that it must contain a 32-bit address. So regardless of the size of the containing field in the note header, a little endian architecture will store the 32-bit address in the first 4 bytes of the buffer and I changed the code to retrieve that value.
Please let me know if you see any issues with this approach.

src/loader/elf.rs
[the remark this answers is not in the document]
andreeaflorescu: Yap, I manually edited this file because we didn't need all those defines. I am fine with manually editing it actually and only include what we need in it. Thoughts?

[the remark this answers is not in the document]
aghecenco: > I am fine with manually editing it actually and only include what we need in it. 

I'm OK with deleting unnecessary code but not with adding new code which would be lost when regenerating the file.

> perhaps the better solution is to run `bindgen` again on `include/uapi/linux/elf.h`

Agreed.

> but I think that is out of the scope for this specific PR. If you don't agree, please let me know and I'll create a new `elfnote.rs` to contain the Note definitions I am adding.

I'd be OK with leaving them in if the naming was consistent with what `bindgen` produces. IMO you could copy-paste the relevant definitions and tests instead of manually defining them. I ran `bindgen` and it looks like the corresponding struct is `elf64_note`?

>  I did include the` pub const XEN_ELFNOTE_PHYS32_ENTRY: u32 = 18;` definition here that comes from a different header
> That should be put somewhere else since it will be clobbered if bindgen is run again.

Agreed.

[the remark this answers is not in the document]
aljimenezb: Please see last revision, specifically the `pvh: Add ELF Note definitions` commit.

I ran `bindgen` again on `include/uapi/linux/elf.h` and did lots of manual editing. The resulting file is mostly the same, with some minor differences that are likely due to me using a different set of options when invoking bindgen, or to changes to bindgen algorithm since the original `elf.rs` was generated. Nothing too disruptive I think. e.g. constants like `EI_MAG0` are now defined using Rust primitive types (`u32`) as opposed to the C-equivalent types (`::std::os::raw::c_uint`) and  the Clone trait is now automatically derived.

I moved the definition of `XEN_ELFNOTE_PHYS32_ENTRY: u32 = 18;` to the `parse_elf_note()` helper since that is the only place that needs it.

PR #13, opened 2020-03-04
Fix the unit tests

rbradford:
Update for latest vm-memory and fix the download location for the kernel.

### DIFF
--- a/src/loader/mod.rs
+++ b/src/loader/mod.rs
@@ -420,7 +420,7 @@ mod test {
     const MEM_SIZE: u64 = 0x1000000;
 
     fn create_guest_mem() -> GuestMemoryMmap {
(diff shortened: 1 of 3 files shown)


src/loader/mod.rs
```diff
@@ -420,7 +420,7 @@ mod test {
     const MEM_SIZE: u64 = 0x1000000;
 
     fn create_guest_mem() -> GuestMemoryMmap {
-        GuestMemoryMmap::new(&[(GuestAddress(0x0), (MEM_SIZE as usize))]).unwrap()
+        GuestMemoryMmap::from_ranges(&[(GuestAddress(0x0), (MEM_SIZE as usize))]).unwrap()
```
aghecenco: `vm-memory` is now published on crates.io - why not update `Cargo.toml` to pick it up from there instead of Github?

rbradford: Done.

Issue #17, opened 2020-03-09
Improve documentation and usage examples


aghecenco:
Following the example of `vm-memory`, `linux-loader` should have extensively documented public-facing APIs, with unit tested examples.

PR #16, opened 2020-03-09
Aarch64 Image format and device tree loading support for linux-loader

crab2313:
I am working on porting cloud-hypervisor to ARM64 platform and this is part of my effort.

PR #23, opened 2020-03-09
Usage examples, doctests & cosmetic tweaks

aghecenco:
#17

### DIFF
--- a/src/cmdline/mod.rs
+++ b/src/cmdline/mod.rs
@@ -12,7 +12,7 @@ use std::fmt;
 use std::result;
 
 /// The error type for command line building operations.
-#[derive(PartialEq, Debug)]
+#[derive(Debug, PartialEq)]
 pub enum Error {
     /// Operation would have resulted in a non-printable ASCII character.
     InvalidAscii,
@@ -30,16 +30,17 @@ impl fmt::Display for Error {
             f,
             "{}",
             match *self {
-                Error::InvalidAscii => "string contains non-printable ASCII character",
-                Error::HasSpace => "string contains a space",
-                Error::HasEquals => "string contains an equals sign",
-                Error::TooLarge => "inserting string would make command line too long",
+                Error::InvalidAscii => "String contains a non-printable ASCII character.",
+                Error::HasSpace => "String contains a space.",
+                Error::HasEquals => "String contains an equals sign.",
+                Error::TooLarge => "Inserting string would make command line too long.",
             }
         )
     }
 }
 
-/// Specialized Result type for command line operations.
+/// Specialized [`Result`] type for command line operations.
+/// [`Result`]: https://doc.rust-lang.org/std/result/enum.Result.html
 pub type Result<T> = result::Result<T, Error>;
 
 fn valid_char(c: char) -> bool {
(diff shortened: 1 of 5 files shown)


src/cmdline/mod.rs
```diff
@@ -12,7 +12,7 @@ use std::fmt;
 use std::result;
 
 /// The error type for command line building operations.
-#[derive(PartialEq, Debug)]
+#[cfg_attr(test, derive(Debug, PartialEq))]
```
andreeaflorescu: Any reason for not having `Debug` and `PartialEq` in release build as well?

aghecenco: Good catch. These are leftovers from playing with `kcov` runs. I'll clean them up.
They will likely be needed when this crate is consumed by a vmm.

src/loader/struct_util.rs
```diff
@@ -7,14 +7,21 @@
 //
 // SPDX-License-Identifier: Apache-2.0 AND BSD-3-Clause
 
+//! Utility functions for struct manipulation.
```
andreeaflorescu: This shouldn't be a public module, we could drop the high level comment (or not, your choice).

src/loader/struct_util.rs
```diff
@@ -24,6 +31,30 @@ pub type Result<T> = std::result::Result<T, Error>;
 /// * `f` - The input to read from.  Often this is a file.
 /// * `out` - The struct to fill with data read from `f`.
 ///
+/// # Examples
```
andreeaflorescu: Since this is not a public module, can we also drop the examples? They are a bit superfluous for functions that only have 2 lines of meaningful code.

src/loader/mod.rs
```diff
@@ -288,20 +319,45 @@ pub struct BzImage;
...
+    ///     &gm,
+    ///     Some(kernel_addr),
+    ///     &mut Cursor::new(&kernel_image),
+    ///     Some(himem_start),
+    /// ).unwrap();
+    /// assert_eq!(loader_result.kernel_load.raw_value(), 0x200000);
+    /// assert_eq!(loader_result.setup_header.unwrap().header, 0x53726448);
+    /// assert_eq!(loader_result.setup_header.unwrap().version. 0x20d);
+    /// assert_eq!(loader_result.setup_header.unwrap().loadflags, 1);
+    /// assert_eq!(loader_result.kernel_end, 0x60c320);
```
andreeaflorescu: Can we drop these asserts? It makes the code example longer than it needs to be. These should be tested with unit test instead (and probably they are).

src/cmdline/mod.rs
```diff
@@ -69,17 +70,37 @@ fn valid_element(s: &str) -> Result<()> {
...
+///
+/// # Examples
+///
+/// ```rust
+/// # use linux_loader::cmdline::*;
+/// # use std::ffi::CString;
+/// let cl = Cmdline::new(100);
+/// let cl_cstring = CString::new(cl).unwrap();
+/// assert_eq!(cl_cstring.to_str().unwrap(), "");
+/// ```
```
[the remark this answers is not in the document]
aghecenco: It doesn't appear to be needed anywhere, anyway.

src/loader/mod.rs
aghecenco: Yup

src/loader/struct_util.rs
aghecenco: Same as above, I think it should be public, but in another crate.

andreeaflorescu: I don't really mind it being public in another crate, but I don't really see the point in the linux loader exporting a `struct_util` module that doesn't really have anything to do with loading kernels. How about for now we keep it private and create a PR in vmm-sys-util instead with this utility and its documentation?

aghecenco: That's where I was aiming too, this would have been an intermediary step but I agree with skipping it and going straight to `vmm-sys-util`.
BTW, `cmdline` is in the same situation. There's no point for `linux-loader` to publicly export functions that compose strings from key-value pairs.

src/loader/struct_util.rs
iulianbarbu: I think we should wait for #26 to be merged. This wouldn't represent a concern anymore.

src/lib.rs
```diff
@@ -11,20 +11,18 @@
...
 //!
 //! # Platform support
 //!
-//! - x86_64
+//! - `x86_64`
 //!
-//! This crates only supports x86_64 platforms because it implements support
-//! for kernel image formats (vmlinux and bzImage) that are x86 specific.
+//! Extending this crate to support other kernel image formats (e.g. ARM's Image) will make it
+//! consumable by other platforms.
```
sameo: I'd rather keep the documentation current and modify it as part of PR #16

sameo: If anything we could add the PVH support to the list.

PR #24, opened 2020-03-10
Reorganize code into x86_64/{elf, bzimage} and aarch64 modules

aghecenco:
The first 2 commits in this PR split the code that currently bloats `src/loader/mod.rs` into:
1. dedicated `elf` and `bzimage` modules under `loader` 
1. per-arch modules:
   1. `x86_64/elf`
   1. `x86_64/bzimage`
   1. `aarch64`

Overall, the purpose of this reorganization is to improve readability and maintainability by keeping `src/loader/mod.rs` simple.


src/loader/x86_64/elf/mod.rs
```diff
@@ -0,0 +1,468 @@
...
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        write!(f, "Kernel Loader Error: {}", self.description())
+    }
+}
+
+/// Raw ELF (a.k.a. vmlinux) kernel image support.
+pub struct Elf;
+
+impl Elf {
+    fn validate_header(ehdr: &elf::Elf64_Ehdr) -> std::result::Result<(), Error> {
```
iulianbarbu: Should we add some documentation to this method?

src/loader/x86_64/elf/mod.rs
```diff
@@ -0,0 +1,468 @@
...
+        }
+
+        // elf image has no setup_header which is defined for bzImage
+        loader_result.setup_header = None;
+
+        Ok(loader_result)
+    }
+}
+
+fn parse_elf_note<F>(phdr: &elf::Elf64_Phdr, kernel_image: &mut F) -> Result<Option<GuestAddress>>
```
iulianbarbu: This function was not documented before. Would it make sense to add some documentation?

aghecenco:
> Looks good to me. I have only one question. Is `start_info.rs` needed ? It doesn't seem to be used inside the crate.

https://github.com/rust-vmm/linux-loader/pull/11#discussion_r391036810 explains why this crate should export the structs in `start_info` and how the VMM that imports `linux-loader` should use them.

src/loader/x86_64/elf/mod.rs
aghecenco: Thanks for catching these 2, I'll take the opportunity to document them. They're not public facing so I'll skip doctests but I agree that they could use some documentation :+1:

aljimenezb: This one is on me :). The description for `parse_elf_note()` should be something like:

"Examines a supplied ELF program header of type PT_NOTE to determine if it contains an entry of type XEN_ELFNOTE_PHYS32_ENTRY (0x12). Notes of this type encode a physical 32-bit entry point address into the kernel which is used when launching guests in 32-bit (protected) mode with paging disabled, as described by the PVH boot protocol. 

On a successful execution, parse_elf_note() returns an optional type containing the encoded entry point address, or None if no entries of the XEN_ELFNOTE_PHYS32_ENTRY type are found in the note header provided as the first argument."

Or something to that effect..

aghecenco: Thanks @aljimenezb, I added the documentation.

andreeaflorescu:
nit: Can we organize the exports as they're organized in vmm-sys-util? You can check [src/lib.rs](https://github.com/rust-vmm/vmm-sys-util/blob/master/src/lib.rs) and the mod.rs file from unix. The advantage with that approach is that when you add a new feature you don't have to grow the lib.rs file and have both the attributes for the features & for the platform at the same level.

aghecenco:
> nit: Can we organize the exports as they're organized in vmm-sys-util? You can check [src/lib.rs](https://github.com/rust-vmm/vmm-sys-util/blob/master/src/lib.rs) and the mod.rs file from unix. The advantage with that approach is that when you add a new feature you don't have to grow the lib.rs file and have both the attributes for the features & for the platform at the same level.

I don't understand which structure you have in mind for `lib.rs`. It only exports `pub mod loader`, so it's pretty small. `loader/mod.rs` is more complicated because the error and `KernelLoaderResult` types there aggregates errors from all the other `mod`s... I can't figure out a way to simplify the imports.

aghecenco:
@rbradford / @sameo / @sboeuf Could either of you PTAL too?

src/loader/mod.rs
```diff
@@ -39,110 +33,61 @@ use vm_memory::{Address, Bytes, GuestAddress, GuestMemory, GuestUsize};
...
+
+#[cfg(all(feature = "bzimage", any(target_arch = "x86", target_arch = "x86_64")))]
+pub use x86_64::bzimage::BzImage;
+#[cfg(all(feature = "bzimage", any(target_arch = "x86", target_arch = "x86_64")))]
+pub use x86_64::bzimage::Error as BzImageError;
+
+#[cfg(all(feature = "elf", any(target_arch = "x86", target_arch = "x86_64")))]
+pub use x86_64::elf::Elf;
+#[cfg(all(feature = "elf", any(target_arch = "x86", target_arch = "x86_64")))]
+pub use x86_64::elf::Error as ElfError;
```
sameo: I think you also need a:
```Rust
#[cfg(all(feature = "elf", any(target_arch = "x86", target_arch = "x86_64")))]
pub use x86_64::elf::start_info;
```

aghecenco: I think (hope) both this issue and @andreeaflorescu's nit https://github.com/rust-vmm/linux-loader/pull/24#issuecomment-600578026 are solved in the latest commit. The `elf` and `bzimage` modules are `use`d as-is, simplifying the imports in `loader/mod.rs`, at the expense of the user prepending `elf::` and `bzimage::` to structs and calls.

aghecenco:
@iulianbarbu @serban300 @sameo @andreeaflorescu Lost all approvals, rebased & updated, PTAL :smiley:

sameo:
Works with Cloud Hypervisor.

src/loader/elf/mod.rs
```diff
@@ -0,0 +1,470 @@
...
+    /// # use linux_loader::loader::*;
+    /// # use vm_memory::{Address, GuestAddress, GuestMemoryMmap};
+    /// # use std::io::Cursor;
+    /// let mem_size: usize = 0x1000000;
+    /// let himem_start = GuestAddress(0x0);
+    /// let kernel_addr = GuestAddress(0x200000);
+    /// let gm = GuestMemoryMmap::from_ranges(&[(GuestAddress(0x0), mem_size)]).unwrap();
+    /// let mut kernel_image = vec![];
+    /// kernel_image.extend_from_slice(include_bytes!("test_elf.bin"));
+    /// assert!(Elf::load(
```
andreeaflorescu: Same comment about replacing asserts with unwraps.

src/loader/mod.rs
```diff
@@ -150,12 +95,26 @@ impl Display for Error {
...
+        Error::Bzimage(err)
+    }
+}
+
+/// Result of [`KernelLoader.load()`](trait.KernelLoader.html#tymethod.load).
 ///
 /// This specifies where the kernel is loading and passes additional
 /// information for the rest of the boot process to be completed by
 /// the VMM.
+#[derive(Default)]
```
andreeaflorescu: Can we keep all the derives? They are actually pretty nice when debugging the code.

src/loader/bzimage/mod.rs
```diff
@@ -0,0 +1,246 @@
...
+    /// # use linux_loader::loader::*;
+    /// # use vm_memory::{Address, GuestAddress, GuestMemoryMmap};
+    /// # use std::io::Cursor;
+    /// let mem_size: usize = 0x1000000;
+    /// let himem_start = GuestAddress(0x0);
+    /// let kernel_addr = GuestAddress(0x200000);
+    /// let gm = GuestMemoryMmap::from_ranges(&[(GuestAddress(0x0), mem_size)]).unwrap();
+    /// let mut kernel_image = vec![];
+    /// kernel_image.extend_from_slice(include_bytes!("bzimage"));
+    /// assert!(BzImage::load(
```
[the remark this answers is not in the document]
aghecenco: Fixed in the latest version for both occurrences. Check the commit that splits the codebase into `x86_64` and `aarch64`.

src/loader/mod.rs
aghecenco: Done

aghecenco:
> @iulianbarbu @serban300 @sameo @andreeaflorescu Lost all approvals, rebased & updated, PTAL

Lost again but I hope this is the last time I ask for re-approvals

sboeuf:
>Lost again but I hope this is the last time I ask for re-approvals

Do you think we could disable this feature? I mean unless we really don't trust the people submitting code, I don't expect someone to shove some bad code on purpose between two iterations of a PR.
Moreover, if someone feel like the PR is not ready while someone else already approved it, it's only a matter of clicking the `Request Changes` button when doing the review. This should prevent the PR from being merged.
What do y'all think?

andreeaflorescu:
@sboeuf I am a bit on the fence about this one. The thing is that you can make mistakes on rebases so someone should still look at the code to make sure you didn't introduce any bugs.

Ideally, you would need to re-approve it only when the PR has changes that were not rebase specific. GitHub helps with that by allowing you to click the `update branch` button that does rebases for you while keeping the approves. The problem is that with PRs that sit a lot in the queue before getting merged, you will have conflicts that GitHub doesn't know how to fix. The same happens for PRs that change lots of lines of code (like this one). 

How about before taking a decision, we put some metrics in place to see how often it happens that we need re-approvals where they would not be necessary and take a data-driven decision? I personally don't feel comfortable with taking that decision now.

I would also move this conversation either in a community issue or on the email so that it doesn't get lost when this PR is merged.

sboeuf:
@andreeaflorescu 
>The thing is that you can make mistakes on rebases so someone should still look at the code to make sure you didn't introduce any bugs.

That's why we put our trust in the CI :)

>How about before taking a decision, we put some metrics in place to see how often it happens that we need re-approvals where they would not be necessary and take a data-driven decision? I personally don't feel comfortable with taking that decision now.

Of course, there's no rush, but that's something I find a bit heavy in terms of process, and this PR is the perfect example, that's why I wanted to raise the question here.

>I would also move this conversation either in a community issue or on the email so that it doesn't get lost when this PR is merged.

Sounds good!

PR #28, opened 2020-03-19
build: Bump the minimum version of vm-memory

rbradford:
"0.1.0" means every version >= 0.1.0 < 0.2.0 but this breaks as the
latest vm-memory release is 0.2.0.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -11,4 +11,4 @@ elf = []
 bzimage = []
 
 [dependencies]
-vm-memory = {version = "0.1.0", features = ["backend-mmap"]}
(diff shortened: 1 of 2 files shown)


andreeaflorescu:
Can we do this update & remove struct-util in the same commit?

bonzini:
Why not `0, >=0.2.0`? I don't think we expect any breaking changes that would bump only the minor version.

rbradford:
> Why not `0, >=0.2.0`? I don't think we expect any breaking changes that would bump only the minor version.

Yes good idea. I think @andreeaflorescu is right this should be done as part of #26 and it should follow your advice to be `^0, >=0.2.0`

PR #29, opened 2020-03-22
Fix possible bugs when loading ELF/PVH kernel 

jiangliu:
This series includes:
1) a patch to exactly matching PVH entry point in ELF note section.
2) do not assume ELF program headers are sorted ascendantly by virtual address.
3) refine the dependency relationship to vm-memory.
Thanks!


sameo:
This LGTM but I'd like to get @aljimenezb review on that one.

src/loader/mod.rs
```diff
@@ -293,10 +293,11 @@ impl KernelLoader for Elf {
                 .read_exact_from(mem_offset, kernel_image, phdr.p_filesz as usize)
                 .map_err(|_| Error::ReadKernelImage)?;
 
-            loader_result.kernel_end = mem_offset
+            let kernel_end = mem_offset
```
bjzhjing: Is there a specific type of segment will have such issue? Refer to [elf](https://refspecs.linuxbase.org/elf/elf.pdf), it says "Loadable segment entries in the program header table appear in ascending order, sorted on the p_vaddr member" under PT_LOAD.

andreeaflorescu:
@jiangliu is it possible to also add unit tests for the bug fixes? I strongly encourage to have regression tests whenever we fix bugs to make sure we don't introduce them again.

src/loader/mod.rs
jiangliu: There's no concrete example, just for safety:)

src/loader/mod.rs
```diff
@@ -306,6 +306,11 @@ impl KernelLoader for Elf {
     }
 }
 
+#[cfg(feature = "elf")]
+#[cfg(any(target_arch = "x86", target_arch = "x86_64"))]
+// Size of string "PVHNote", including the terminating NULL.
+const PVH_NOTE_STR_SZ: usize = 8;
+
```
aljimenezb: I agree that checking for both name and type when searching for the PVH note is the right approach, for the reason @jiangliu mentions in the commit message. This will avoid any future collisions due to another namespace using type 0x12 as well.

This patch will not work correctly yet, because the name specified in the [PVH ABI](https://xenbits.xen.org/docs/unstable/misc/pvh.html) is actually just "Xen". So the code above must be changed to:
```rust
// Size of string "Xen", including the terminating NULL.
const PVH_NOTE_STR_SZ: usize = 4;
```
The unit tests I added on [this commit](https://github.com/rust-vmm/linux-loader/commit/0ce5bfa70d1073a9743fadfea233edef09c81e49) use "PVHNote" as the name to generate the test_*note.bin binaries and they must be corrected. I can generate new binaries and share them, or they can be built using the instructions in the commit. Please let me know if you'd like me to do this.

src/loader/mod.rs
```diff
@@ -336,9 +341,16 @@ where
...
         // If the note header found is not the desired one, keep reading until
         // the end of the segment
-        if nhdr.n_type == XEN_ELFNOTE_PHYS32_ENTRY {
-            break;
+        if nhdr.n_type == XEN_ELFNOTE_PHYS32_ENTRY && nhdr.n_namesz as usize == PVH_NOTE_STR_SZ {
+            let mut buf = [0u8; PVH_NOTE_STR_SZ];
+            kernel_image
+                .read_exact(&mut buf)
+                .map_err(|_| Error::ReadNoteHeader)?;
+            if buf == [b'P', b'V', b'H', b'N', b'o', b't', b'e', b'\0'] {
```
aljimenezb:
```rust
if buf == [b'X', b'e', b'n', b'\0'] {
```
is needed here for the reason mentioned above.

src/loader/mod.rs
aljimenezb: @jiangliu I built new binaries that are needed to fix the unit tests. Binaries, source and instructions on how to build are here:
https://gist.github.com/aljimenezb/9918b8562070cb89ab6696025305e3f9

jiangliu: Yes, the linux kernel use "Xen" and I used "Xen" too at first, but it fails the unit tests:(
I will update it in next version.

andreeaflorescu: @aljimenezb can you add the instructions about generating the binaries to this repository as well? It would be easier if someone else needs to generate some in the future.

aljimenezb: @andreeaflorescu Sure, I'll open a separate PR for that change. That content feels a little bit out of place in the main README, but we can figure out the details on the discussion.

andreeaflorescu: Yeah, maybe we can have a contributing document? Or development?

Another idea would be for those binaries to be generated at build time. I am not a big fun of having binaries in repositories.

aljimenezb: > Another idea would be for those binaries to be generated at build time. I am not a big fun of having binaries in repositories.

That was my first instinct, but I saw test_elf.bin was already in use I followed that simpler approach instead.

 I am not familiar with the capabilities of inline assembly in Rust, so I'd like feedback from others here as to whether or not it is possible to build the [assembly source code](https://gist.github.com/aljimenezb/9918b8562070cb89ab6696025305e3f9) using Rust inline assembly templates.

aljimenezb:
> 
> 
> This PR breaks PVH boot (the code goes back through the vmlinux) path when it was tested with Cloud Hypervisor. I recommend you put a debug statement in `arch::x86_64::configure_system` to print out the value of `boot_proto`.

@rbradford I am assuming you tested with the current code, and without the changes I suggested above. In that case yes, it will break PVH boot since the Linux kernel uses "Xen" as the name for the note header that encodes the PVH address. 

After applying my proposed changes above, I booted successfully using the PVH path on Cloud Hypervisor. If you did the same and still found the error please let me know and I'll take another look.

rbradford:
> @rbradford I am assuming you tested with the current code, and without the changes I suggested above. In that case yes, it will break PVH boot since the Linux kernel uses "Xen" as the name for the note header that encodes the PVH address.

Indeed.

sboeuf:
@jiangliu just wanted to mention that I tried this PR yesterday, and it breaks the PVH support.

jiangliu:
> @jiangliu just wanted to mention that I tried this PR yesterday, and it breaks the PVH support.

Yes, the patch should be updated to checking "Xen" instead of "PVHNote".
Will send out updated version soon.

src/loader/x86_64/elf/mod.rs
```diff
@@ -281,11 +284,18 @@ where
             .read_from(0, kernel_image, nhdr_sz)
             .map_err(|_| Error::ReadNoteHeader)?;
 
-        // If the note header found is not the desired one, keep reading until the end of the
-        // segment.
-        if nhdr.n_type == XEN_ELFNOTE_PHYS32_ENTRY {
-            break;
+        // If the note header found is not the desired one, keep reading until
+        // the end of the segment
```
aljimenezb: I would change this comment since you are adding checks other than the very simple one it had before. Perhaps something like:
``// Check if the note header's name and type match the ones specified by the PVH ABI.``

src/loader/x86_64/elf/mod.rs
```diff
@@ -281,11 +284,18 @@ where
...
+        if nhdr.n_type == XEN_ELFNOTE_PHYS32_ENTRY && nhdr.n_namesz as usize == PVH_NOTE_STR_SZ {
+            let mut buf = [0u8; PVH_NOTE_STR_SZ];
+            kernel_image
+                .read_exact(&mut buf)
+                .map_err(|_| Error::ReadNoteHeader)?;
+            if buf == [b'x', b'e', b'n', b'\0'] {
```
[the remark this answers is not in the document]
aljimenezb: I see you corrected this on the [next commit](https://github.com/rust-vmm/linux-loader/pull/29/commits/025db34b64039a76d80439dcebed3d5d0602e968). Why not just fix it on the commit that introduces the code?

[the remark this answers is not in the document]
jiangliu: sorry, a mistake when merging patch series.

josephlr:
rust-hypervisor-firmware recently added PVH support in https://github.com/cloud-hypervisor/rust-hypervisor-firmware/pull/26.

I tested this patch w/ Cloud-Hypervisor, and I can confirm that this patch doesn't break anything with regards to the firmware. We can still boot via the PVH Boot protocol.

Issue #15, opened 2020-03-05
Take care of writing boot parameters into memory


bonzini:
The linux-loader crate does not handle writing in memory the setup header that it prepares, and there could also be extra parameters (e.g. kernel command line or, for PVH, the e820 map) that need to be written into memory. I am therefore proposing to add a "configurator" object that encapsulates all the logic needed to boot a kernel (later on this could include the CPU registers too). The basic structure would be:

```rust
pub struct E820MapEntry { ... |}
pub struct BootParameters {
    e820: Vec<E820MapEntry>, ...
};
...
pub trait BootConfigurator<M: GuestMemory> {
    pub fn write_boot_parameters(&self, result: &KernelLoaderResult, params: &BootParameters, dest: T);
}
pub struct PVHConfigurator; // implements BootConfigurator<T>
pub struct LinuxElfConfigurator; // implements BootConfigurator<T>
pub struct BzImageConfigurator; // implements BootConfigurator<T>
```

and probably an associated type on KernelLoader like

```rust
pub trait KernelLoader<M: GuestMemory> {
    type Configurator: Box<dyn BootConfigurator<M>>;
}

pub struct KernelLoaderResult<M: GuestMemory> {
    ...
    configurator: Box<dyn BootConfigurator<M>>
}

```

that can be used like

```rust
    let result = KernelLoaderType.load(guest_mem, ...)?;
    result.configurator.write_boot_parameters(result, guest_mem);
```

andreeaflorescu:
Yap, it makes a lot of sense. For example, in Firecracker we have this logic in the `arch` crate and I think it shouldn't be there. This proposal looks good to me.

sameo:
@bonzini I guess you meant:

```Rust
let result = KernelLoaderType.load(guest_mem, ...)?;
let boot_params = arch::build_boot_params()?;
result.configurator.write_boot_parameters(result, boot_params, guest_mem);
```

where the `linux_loader` boot parameters would be built from the VMM?

bonzini:
Yes, precisely.

aghecenco:
After looking more closely at the code for *building* bootparams (at least for `elf`), there's a lot of it in common. I think that can be packaged too in a builder of sorts, extended in the `vmm` if needed, then passed on to the `Configurator`.

```rust
pub trait BootConfigurator {
    type Error;

    fn write_bootparams<B: ByteValued, M: GuestMemory>(
        params: B,
        guest_memory: &M,
[...]

PR #31, opened 2020-03-31
Boot configurator: trait (and associated objects) that write boot params in guest memory

aghecenco:
#15 

This PR introduces a new `configurator` module, meant to build (TBD) and write (included in this PR) boot parameters in guest memory. To begin with, the VMM will have to build its own boot params, encapsulate them into [`ByteValued`](https://docs.rs/vm-memory/0.2.0/vm_memory/bytes/trait.ByteValued.html) objects, then pass them on to `linux-loader`.

VMM workflow:

```rust
let load_result = kernel_loader.load<ReadImpl, GuestMemoryImpl>(
        guest_memory,
        kernel_start_addr,
        kernel_image,
        himem_start_addr
)?;
let (boot_params_header, boot_params_sections): (HeaderImpl, Vec<SectionImpl>) = Vmm::build_boot_params()?;
BootConfiguratorImpl::write_bootparams::<HeaderImpl, SectionImpl, GuestMemoryImpl>(
        (boot_params_header, header_addr),
        Some((boot_params_sections, sections_addr)),
        guest_memory
)?;
```

The PR adds support for 3 boot protocols:

### `x86_64` Linux boot
* `header`:  [`boot_params`](https://github.com/rust-vmm/linux-loader/blob/2adddce25b37f68c99ae8e7db1cb7f5853626fdd/src/loader/bootparam.rs#L3475) struct, everything that goes into the zeropage
* `sections`: unused (e820 map included in `header`)

```rust
let params: boot_params = Vmm::build_boot_params_linux()?;
LinuxBootConfigurator::write_bootparams::<boot_params, boot_params, GuestMemoryImpl>(
        (params, zeropg_addr),
        None,
        guest_memory
)?;
```

### `x86_64` PVH boot
* `header`:  [`hvm_start_info`](https://github.com/rust-vmm/linux-loader/blob/2adddce25b37f68c99ae8e7db1cb7f5853626fdd/src/loader/x86_64/elf/start_info.rs#L126) struct
* `sections`: vector of [`hvm_memmap_table_entry`](https://github.com/rust-vmm/linux-loader/blob/2adddce25b37f68c99ae8e7db1cb7f5853626fdd/src/loader/x86_64/elf/start_info.rs#L150) structs: memory map table entries

```rust
let (start_info, memmap_entries): (hvm_start_info, Vec<hvm_memmap_table_entry>) = Vmm::build_boot_params_pvh()?;
PvhBootConfigurator::write_bootparams::<hvm_start_info, hvm_memmap_table_entry, GuestMemoryImpl>(
        (start_info, start_info_addr),
        Some((memmap_entries, memmap_addr)),
        guest_memory
)?;
```

### `aarch64` boot

* `header`: byte blob containing the compiled FDT
* `sections`: unused

```rust
struct FdtBlob { ... }
impl ByteValued for FdtBlob { ... }
...
let fdt_blob: FdtBlob= Vmm::build_fdt()?;
FdtBootConfigurator::write_bootparams::<FdtBlob, FdtBlob, GuestMemoryImpl>(
        (fdt_blob, fdt_addr),
        None,
        guest_memory
)?;
```

### DIFF
--- a/src/configurator/aarch64/fdt.rs
+++ b/src/configurator/aarch64/fdt.rs
@@ -59,7 +59,7 @@ impl BootConfigurator for FdtBootConfigurator {
     {
         // The VMM has filled an FDT and passed it as a `ByteValued` object.
         guest_memory
-            .write_slice(params.header.0.as_slice(), params.header.1)
+            .write_slice(params.header.as_slice(), params.header_start)
             .map_err(|_| Error::WriteFDTToMemory.into())
     }
 }
--- a/src/configurator/aarch64/fdt.rs
+++ b/src/configurator/aarch64/fdt.rs
@@ -1,0 +1,72 @@
+// Copyright 2020 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+//
+// SPDX-License-Identifier: Apache-2.0 AND BSD-3-Clause
+
+//! Traits and structs for loading the device tree.
+
+use vm_memory::{ByteValued, Bytes, GuestAddress, GuestMemory};
+
+use std::error::Error as StdError;
+use std::fmt;
+
+use super::super::{BootConfigurator, Error as BootConfiguratorError, Result};
+
+/// Errors specific to the device tree boot protocol configuration.
+#[derive(Debug, PartialEq)]
+pub enum Error {
+    /// Error writing FDT in memory.
+    WriteFDTToMemory,
+}
+
+impl StdError for Error {
+    fn description(&self) -> &str {
+        use Error::*;
+        match self {
+            WriteFDTToMemory => "Error writing FDT in guest memory.",
+        }
+    }
+}
+
+impl fmt::Display for Error {
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        write!(
+            f,
+            "Device Tree Boot Configurator Error: {}",
+            StdError::description(self)
+        )
+    }
+}
+
+impl From<Error> for BootConfiguratorError {
+    fn from(err: Error) -> Self {
+        BootConfiguratorError::Fdt(err)
+    }
+}
+
+/// Boot configurator for device tree.
+pub struct FdtBootConfigurator {}
+
+impl BootConfigurator for FdtBootConfigurator {
+    /// Writes the boot parameters (configured elsewhere) into guest memory.
+    ///
+    /// # Arguments
+    ///
+    /// * `fdt` - flattened device tree.
+    /// * `sections` - unused.
+    /// * `guest_memory` - guest's physical memory.
+    fn write_bootparams<T, S, M>(
+        fdt: (T, GuestAddress),
+        _sections: Option<(Vec<S>, GuestAddress)>,
+        guest_memory: &M,
+    ) -> Result<()>
+    where
+        T: ByteValued,
+        S: ByteValued,
+        M: GuestMemory,
+    {
+        // The VMM has filled an FDT and passed it as a `ByteValued` object.
+        guest_memory
+            .write_obj(fdt.0, fdt.1)
+            .map_err(|_| Error::WriteFDTToMemory.into())
+    }
+}
--- a/src/configurator/aarch64/fdt.rs
+++ b/src/configurator/aarch64/fdt.rs
@@ -14,6 +14,8 @@ use crate::configurator::{BootConfigurator, BootParams, Error as BootConfigurato
 /// Errors specific to the device tree boot protocol configuration.
 #[derive(Debug, PartialEq)]
 pub enum Error {
+    /// FDT does not fit in guest memory.
+    FDTPastRamEnd,
     /// Error writing FDT in memory.
     WriteFDTToMemory,
 }
@@ -22,6 +24,7 @@ impl StdError for Error {
     fn description(&self) -> &str {
         use Error::*;
         match self {
+            FDTPastRamEnd => "FDT does not fit in guest memory.",
             WriteFDTToMemory => "Error writing FDT in guest memory.",
         }
     }
@@ -57,6 +60,10 @@ impl BootConfigurator for FdtBootConfigurator {
     where
         M: GuestMemory,
     {
+        guest_memory
+            .checked_offset(params.header_start, params.header.len())
+            .ok_or(Error::FDTPastRamEnd)?;
+
         // The VMM has filled an FDT and passed it as a `ByteValued` object.
         guest_memory
             .write_slice(params.header.as_slice(), params.header_start)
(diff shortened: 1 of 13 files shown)


aghecenco:
Coverage test fails, but I'd like to get some preliminary feedback before investing the extra effort into additional unit tests.

src/configurator/x86_64/pvh.rs
```diff
@@ -0,0 +1,238 @@
...
+unsafe impl ByteValued for hvm_start_info {}
+unsafe impl ByteValued for hvm_memmap_table_entry {}
+
+impl BootConfigurator for PvhBootConfigurator {
+    /// Writes the boot parameters (configured elsewhere) into guest memory.
+    ///
+    /// # Arguments
+    ///
+    /// * `header` - boot parameters encapsulated in a [`hvm_start_info`] struct.
+    /// * `sections` - memory map table represented as a vector of [`hvm_memmap_table_entry`].
```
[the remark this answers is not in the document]
aghecenco: Thank you for the detailed explanation! I updated this PR with some more abstractions over the tuples I initially used, and now the `BootConfigurator` works with a `BootParams` object that has a (mandatory) header (`boot_params` / `hvm_start_info` / fdt blob), optional sections (PVH memmap) and optional modules (PVH modlist for initrd).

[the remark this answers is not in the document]
bonzini: Yes, I was also going to remark that the boot parameters should all be encapsulated in a single struct.

aghecenco:
I used this PR as a starting point to play around with encapsulating bootparams *creation* too, towards the vision of having all kernel loading logic in this crate. I started with elf + Linux as it's the easiest of combinations (building the fdt for ARM is, I think, beyond this crate).

The code is in this branch: https://github.com/aghecenco/linux-loader/tree/build_params

elf + Linux boot: [`LinuxBootParams`](https://github.com/aghecenco/linux-loader/blob/71ad3d9177c66fab1e498e47c53b405efaa73f18/src/configurator/x86_64/linux.rs#L123)

Tried it out with Firecracker: https://github.com/aghecenco/firecracker/tree/linloader
[`configure_system`](https://github.com/aghecenco/firecracker/blob/7d68785f4cafd963e1ac4605776335729b40be7f/src/arch/src/x86_64/mod.rs#L91) is simplified

And cloud-hypervisor: https://github.com/aghecenco/cloud-hypervisor/tree/linloader
[`configure_64bit_boot`](https://github.com/aghecenco/cloud-hypervisor/blob/7f9dcc214d5b333421b2c7f0bbf31c5ba3445bd9/arch/src/x86_64/mod.rs#L304) is simplified

src/configurator/aarch64/fdt.rs
```diff
@@ -0,0 +1,127 @@
...
+        let guest_memory = create_guest_mem();
+
+        // Error case: FDT doesn't fit in guest memory.
+        let fdt_addr = guest_memory
+            .last_addr()
+            .checked_sub(FDT_MAX_SIZE as u64 - 2)
+            .unwrap();
+        assert_eq!(
+            FdtBootConfigurator::write_bootparams::<
+                FdtPlaceholder,
```
aghecenco, line 98: Note to self: to avoid this level of verbosity, maybe `impl ByteValued for () {}` in `vm-memory`

aghecenco:
@bonzini the latest version uses a single `BootParams` object with (up to) 3 generics for the various components.
@aljimenezb the `modlist` is an `Option`al member of the `BootParams`, filled in only for PVH.
Could you please take another look?

aljimenezb:
@aghecenco Sorry it took me a bit to reply back. The new types look good to me, but probably others with more Rust experience should take a look. One minor suggestion is to avoid having members of `BootParams` and `BootSections` that are both called `sections`, to potentially avoid lines like:

 `let memmap_list = params.sections.unwrap().sections;`

Perhaps the second field in the `BootParams` struct can be renamed to `mem_maps` or something similar, since it is only used for PVH to specify memory maps anyway. 

Also I noticed that the write_bootparams() implementation in PvhBootConfigurator did not have support yet for copying the module list into guest memory, so I added the (untested!) code here:
https://github.com/aljimenezb/linux-loader/commit/9f932dca5862b989ac20db95239f2acdd00257ed

bonzini:
I think adding more generic parameters is not the right approach.

BootParams can be kept generic if you make it a convenience for building a `Vec<(Box<[u8]>, GuestAddress)>`. It can have generic methods `set_header(&mut self, ...)` `add_section(&mut self, ...)` `add_module(&mut self, ...)` that take any ByteValued and make a copy into a `Box<[u8]>` (I think `vec![0; size].into_boxed_slice()` works for that). Then `write_bootparams` does not need to be generic at all because it just iterates along the `Vec<(Box<[u8]>, GuestAddress)>` and writes each slice to the guest memory.

aghecenco:
@bonzini I got rid of the 3 generics; the struct still separates `header` from `sections` and `modules` to allow for specific errors when writing the params to guest memory.
@alexandruag please take a(nother) look too.

bonzini:
Nice! My only remaining qualm is that add_sections actually sets the whole Vec, so I would prefer an add_section method that can be called multiple times. Likewise for modules.

bonzini:
There can still be add_sections and add_modules methods, taking a Vec, for convenience. However it's a bit limiting to assume a single type for all sections/modules. add_section and add_module could return the top address for the section and module to simplify adding heterogenous sections/modules.

bonzini:
Looks great, thanks!

aghecenco:
I added a new commit that replaces the tuples in the original `BootParams` struct with individual fields, for increased code clarity. I kept the `set_modules` and `set_sections` (very convenient when we already have them) but extended `BootParams` to allow additions of individual sections / modules of various types (as long as they're `ByteValued`) at any valid address. Examples in `configurator/mod.rs` should show how the functions are to be used. Some responsibility is left to the caller, as individual modules and sections can now be placed anywhere after the starting address, potentially overlapping others, or leaving gaps. The caller should use the returned `GuestAddress`es to make sure everything is where they want it to be.

@bonzini many thanks for all the feedback so far, this PR has come a long way since the beginning. Can you take another look? The 3 types of configurators haven't changed much, but the sections/modules rework has built up.

aghecenco:
Hi @bonzini @alexandruag, could you please take a(nother) look?

bonzini:
I don't have approval privileges, but I like the basic API and tests are comprehensive, so it looks good.

src/configurator/aarch64/fdt.rs
```diff
@@ -0,0 +1,120 @@
...
+    /// # Arguments
+    ///
+    /// * `params` - boot parameters containing the FDT.
+    /// * `guest_memory` - guest's physical memory.
+    fn write_bootparams<M>(params: BootParams, guest_memory: &M) -> Result<()>
+    where
+        M: GuestMemory,
+    {
+        // The VMM has filled an FDT and passed it as a `ByteValued` object.
+        guest_memory
```
jiangliu, line 61: Do we need to validate guest memory address here?

aghecenco: Added validation. `write_slice` would have returned an error, but the other use cases explicitly validate that the bootparams fit in guest memory, so I added the same here.

PR #38, opened 2020-06-10
Clarify uses of kernel_start parameter

### LINKED ISSUE #12: Setting load address doesn't work for PVH loader
Setting the kernel load address doesn't result in the kernel booting (tested with Cloud Hypervisor and setting it to 1MiB.) bzImage loader works fine with the specified load address.

aljimenezb:
As part of the discussion in #12, we have noticed that the `kernel_start` parameter has different usages in the implementations for `PE`, `Elf`, and `BzImage`. The `Elf` implementation is specially confusing, since `kernel_start` is used to specify an offset to add to the default load address of the kernel. The discussion is ongoing to determine whether this has any real use cases or provides any security benefits.

In the meantime, I propose to rename the parameter to `kernel_offset`, which might be more appropriate to describe its usage by the implementers of `KernelLoader`.

Also, since PVH code expects the kernel to be loaded at its default load address, it cannot be used to boot guests if a non-default load address is requested by using the `kernel_offset` argument. If `kernel_offset` is requested, we simply avoid looking for a PVH entry point.

Resolves #12

### DIFF
--- a/src/loader/x86_64/elf/mod.rs
+++ b/src/loader/x86_64/elf/mod.rs
@@ -1,3 +1,4 @@
+// Copyright © 2020, Oracle and/or its affiliates.
 // Copyright (c) 2019 Intel Corporation. All rights reserved.
 // Copyright 2018 Amazon.com, Inc. or its affiliates. All Rights Reserved.
 //
@@ -85,6 +86,39 @@ impl error::Error for Error {
     }
 }
 
+#[derive(Clone, Copy, Debug, PartialEq)]
+/// Availability of PVH entry point in the kernel, which allows the VMM
+/// to use the PVH boot protocol to start guests.
+pub enum PvhBootCapability {
+    /// PVH entry point is present
+    PvhEntryPresent(GuestAddress),
+    /// PVH entry point is not present
+    PvhEntryNotPresent,
+    /// PVH entry point is ignored, even if available
+    PvhEntryIgnored,
+}
+
+impl Default for PvhBootCapability {
+    fn default() -> Self {
+        PvhBootCapability::PvhEntryIgnored
+    }
+}
+
+impl Display for PvhBootCapability {
+    fn fmt(&self, f: &mut ::std::fmt::Formatter) -> fmt::Result {
+        use self::PvhBootCapability::*;
+        match self {
+            PvhEntryPresent(pvh_entry_addr) => write!(
+                f,
+                "PVH entry point present at guest address: {:#x}",
+                pvh_entry_addr.raw_value()
+            ),
+            PvhEntryNotPresent => write!(f, "PVH entry point not present"),
+            PvhEntryIgnored => write!(f, "PVH entry point ignored"),
+        }
+    }
+}
+
 impl Display for Error {
     fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
         write!(f, "Kernel Loader Error: {}", self.description())
@@ -121,7 +155,10 @@ impl Elf {
 impl KernelLoader for Elf {
     /// Loads a kernel from a vmlinux elf image into guest memory.
     ///
-    /// The kernel is loaded into guest memory at offset `phdr.p_paddr` specified by the elf image.
+    /// By default, the kernel is loaded into guest memory at offset `phdr.p_paddr` specified
+    /// by the elf image. When used, `kernel_offset` specifies a fixed offset from `phdr.p_paddr`
+    /// at which to load the kernel. If `kernel_offset` is requested, the `pvh_entry_addr` field
+    /// of the result will not be populated.
     ///
     /// # Arguments
     ///
@@ -204,8 +241,17 @@ impl KernelLoader for Elf {
         for phdr in phdrs {
             if phdr.p_type != elf::PT_LOAD || phdr.p_filesz == 0 {
                 if phdr.p_type == elf::PT_NOTE {
-                    // This segment describes a Note, check if PVH entry point is encoded.
-                    loader_result.pvh_entry_addr = parse_elf_note(&phdr, kernel_image)?;
+                    // The PVH boot protocol currently requires that the kernel is loaded at
+                    // the default kernel load address in guest memory (specified at kernel
+                    // build time by the value of CONFIG_PHYSICAL_START). Therefore, only
+                    // attempt to use PVH if an offset from the default load address has not
+                    // been requested using the kernel_offset parameter.
+                    if let Some(_offset) = kernel_offset {
+                        loader_result.pvh_boot_cap = PvhBootCapability::PvhEntryIgnored;
+                    } else {
+                        // If kernel_offset is not requested, check if PVH entry point is present
+                        loader_result.pvh_boot_cap = parse_elf_note(&phdr, kernel_image)?;
+                    }
                 }
                 continue;
             }
@@ -248,9 +294,9 @@ const PVH_NOTE_STR_SZ: usize = 4;
 /// of type `XEN_ELFNOTE_PHYS32_ENTRY` (0x12). Notes of this type encode a physical 32-bit entry
 /// point address into the kernel, which is used when launching guests in 32-bit (protected) mode
 /// with paging disabled, as described by the PVH boot protocol.
-/// Returns the encoded entry point address, or `None` if no `XEN_ELFNOTE_PHYS32_ENTRY` entries are
-/// found in the note header.
-fn parse_elf_note<F>(phdr: &elf::Elf64_Phdr, kernel_image: &mut F) -> Result<Option<GuestAddress>>
+/// Returns the encoded entry point address, or `None` if no `XEN_ELFNOTE_PHYS32_ENTRY` entries
+/// are found in the note header.
+fn parse_elf_note<F>(phdr: &elf::Elf64_Phdr, kernel_image: &mut F) -> Result<PvhBootCapability>
 where
     F: Read + Seek,
 {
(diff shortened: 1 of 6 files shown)


src/loader/x86_64/elf/mod.rs
```diff
@@ -212,7 +213,12 @@ impl KernelLoader for Elf {
         // Read in each section pointed to by the program headers.
         for phdr in phdrs {
             if phdr.p_type != elf::PT_LOAD || phdr.p_filesz == 0 {
-                if phdr.p_type == elf::PT_NOTE {
+                // The PVH boot protocol currently requires that the kernel is loaded at
+                // the default kernel load address in guest memory (specified at kernel
+                // build by the value of CONFIG_PHYSICAL_START). Therefore, only attempt
+                // to use PVH if an offset from the default load address has not been
+                // requested using the kernel_offset parameter.
+                if phdr.p_type == elf::PT_NOTE && kernel_offset.is_none() {
```
andreeaflorescu, line 221: Handling this parameter like so makes for implicit decisions. I find it better in code to let the user have explicit decisions.
With this approach it is easy to mis-use the function and not understand what you did wrong when it happens.

The implicit decision here is:
- if you pass kernel_offset than this must also mean that you don't want PVH even though p_type is PVH

An explicit decision is:
- p_type is PVH => you want to parse PVH

To keep the decision explicit what we can do instead is to ignore kernel_offset for PVH. I looks a bit like this is what was happening before. We can be explicit in the documentation as well and say that for PVH `kernel_offset` is ignored.

What do you say?

aljimenezb: I agree with your point of letting the user make the decisions, but I am not really sure what is the best way to do it in this case. I'm on record saying that `kernel_offset` does not have any real use cases for the `Elf` implementation and it should be an unused parameter, but if we still want to keep it around then it means that PVH can not be returned as an option. The VMM "knows" it is not using PVH, since the return type from the `load()` method has a `pvh_entry_addr` field that is `None`, but it does not know why it can't. 

We can't exactly implement what you suggest above, so let me try to clarify how the PVH boot selection works, and we might be able to come up with a better idea.

The short explanation is that the VMM can not request when it calls `loader::elf::Elf::load()` that PVH is used; it does not know yet if it is available. The availability of PVH is an extra choice, a "bonus feature", that is provided alongside the default boot protocol. But the VMM only knows if it is available after `load()` returns. I didn't provide a way for VMMs to request exclusively that PVH is available or otherwise fail, since that decision can be made at the VMM level.

We cannot ignore the `kernel_offset` request from the VMM and still provide the two options for the VMM to decide, since `kernel_offset` affects how the kernel is loaded in memory. I will add a comment on the parameter description that mentions the effect it has on PVH (it disables it). I didn't want to declare the `kernel_offset/kernel_start` parameter unused in the Elf case, since I can't prove 100% that is not needed.

Hopefully that made sense. Sorry for the long wall of text :).

aljimenezb: Some more details on how the code works, mostly expanding on the previous comment. Feel free to ignore:

The VMM cannot directly request PVH boot, i.e. there is not "p_type is PVH" selection by the VMM. It simply requests to load a kernel that is in the ELF format by calling the `loader::elf::Elf::load()` method.

The ELF header has a field (`e_entry`) which holds the default entry point for the executable. i.e. this is the virtual address that is set in `%rip` by the VMM configuration methods, and the guest will start executing code at this address when it is launched. Before PVH support, this value in `e_entry` was the only entry point that could be used to boot a guest with an ELF (vmlinux) kernel. I'll call this "Linux boot", and this is the way it currently works in Firecracker: 
https://github.com/firecracker-microvm/firecracker/blob/32c5bc3360cb3ee4d836faaa580e5fbfb314ef66/src/kernel/src/loader/mod.rs#L145

The PVH code changes that were merged earlier do not replace the mechanism above, they just add an alternative method, so the VMM now has two choices. Besides returning the default entry point contained in `e_entry`, the load() method now also returns to the VMM an optional field called `pvh_entry_addr`, which if present, tells the VMM that there is another address in the kernel where it can start executing.

src/loader/x86_64/elf/mod.rs
```diff
@@ -130,7 +130,8 @@ impl Elf {
 impl KernelLoader for Elf {
     /// Loads a kernel from a vmlinux elf image into guest memory.
     ///
-    /// The kernel is loaded into guest memory at offset `phdr.p_paddr` specified by the elf image.
+    /// By default, the kernel is loaded into guest memory at offset `phdr.p_paddr` specified by the elf image.
+    /// When used, `kernel_offset` specifies a fixed offset from `phdr.p_paddr` at which to load the kernel.
```
alexandruag, line 134: I think it's useful to explicitly document that whenever `kernel_offset` is present, the `pvh_entry_addr` field of the result will not be populated.

src/loader/x86_64/elf/mod.rs
alexandruag: Apologies if this is too pedantic, but I think we should actually write the `pvh_entry_addr` field of the result if `kernel_offset` is present, but its value is 0.

Another thought was to use an enum with three variants for `pvh_entry_addr` (i.e. `present`/`none`/`ignored`) instead of just an `Option`, but I couldn't think of a scenario where we'd actively look for the distinction, so that probably doesn't add that much value (might be useful for debugging though). WDYT?

src/loader/x86_64/elf/mod.rs
aljimenezb: Agree. Added that line to the comment.

src/loader/x86_64/elf/mod.rs
aljimenezb: > Apologies if this is too pedantic, but I think we should actually write the `pvh_entry_addr` field of the result if `kernel_offset` is present, but its value is 0.
> 
That is a good point that I didn't considered initially. But it seems to me that it contradicts the assertion that if `kernel_offset` is requested, then we do not consider PVH at all. We would be adding an special case in the documented behavior with very little advantage.
That is my current take on it, but if you think that it has any use cases or make things simpler from the VMM side let me know and I'll add it.

> Another thought was to use an enum with three variants for `pvh_entry_addr` (i.e. `present`/`none`/`ignored`) instead of just an `Option`, but I couldn't think of a scenario where we'd actively look for the distinction, so that probably doesn't add that much value (might be useful for debugging though). WDYT?

When adding PVH support I was trying to make minimal changes, and also didn't foresee that `kernel_start` would be a problem, so an `Option` was enough.

The downsides I see with using the enum are:
- It would be a breaking change for VMMs (e.g. changes are needed in Cloud Hypervisor and rust-hypervisor-firmware)
- The `ignored` variant would not be needed if there are future Linux kernel changes to allow PVH to work with `kernel_offset`

Again, I would have to defer to you and others on the discussion more experienced with the VMM implementations as to whether having an enum would make things better for the current main consumer (ICH) and future ones (Firecracker). I would opt to keep it simple and keep using the Option type.

andreeaflorescu: One note on breaking changes: this crate is not even published yet, so I would like to make sure  we have the appropriate interface **before** publishing it. I don't think we have any contract that we will not break the interface before (or even after) publishing. We are working with 0.x.x versions, so we aren't advertising it as stable.

@sameo @rbradford @sboeuf any thoughts on this?

aljimenezb:
Added changes implementing @aghecenco idea of using an `enum` for the PVH field in `KernelLoaderResult`. That is still pending feedback from @sameo @rbradford @sboeuf so I can revert it depending on their comments.

src/loader/x86_64/elf/mod.rs
```diff
@@ -121,7 +154,9 @@ impl Elf {
 impl KernelLoader for Elf {
     /// Loads a kernel from a vmlinux elf image into guest memory.
     ///
-    /// The kernel is loaded into guest memory at offset `phdr.p_paddr` specified by the elf image.
+    /// By default, the kernel is loaded into guest memory at offset `phdr.p_paddr` specified by the elf image.
```
alexandruag, line 157: Just a style comment: can you please reformat these comment lines to keep them under 100 chars?

src/loader/x86_64/elf/mod.rs
alexandruag: Sorry for the late reply. Regarding the different behavior for `kernel_offset == 0`, I was thinking that maybe the value of the offset is specified programmatically in some cases, and when it's 0 the logic can detect that PVH boot is also available as an option and maybe use it. I guess it's just a synthetic thought experiment at this point, so the way things are right now in the PR is also fine.

aljimenezb: I see your point above, but I think the interface is simpler to document and understand for the users of the crate when:
`kernel_offset.is_none()` ==> PVH is not available
`kernel_offset.is_some()` ==> PVH is available
This is predictable and clearly defined. If we add a special case for `kernel_offset == 0`, that blurs the boundaries between those cases. A user of the crate now needs implementation specific knowledge to understand why `kernel_offset == 0` behaves differently.

If the user wants to look into the implementation details then it should be clear enough to understand that `kernel_offset = Some(0)` is the same as `kernel_offset = None` (when it comes to controlling where the kernel is loaded), but you can use PVH with the second variant.

So I'll keep the current approach, since you also mentioned above that you are fine with it. This is also a short change that could be added quickly if the need arises later.

src/loader/x86_64/elf/mod.rs
aljimenezb: Fixed, thanks for the heads up. For some reason my editor (VSCode) does not warn me about those.

aljimenezb:
> Looks good! Can you please add something like `Related to https://github.com/rust-vmm/linux-loader/issues/12` to the message body of each commit? This is a guideline we're trying to implement right now to help with keeping track of things.

Done. Added it right before the Signed-off line.

Issue #35, opened 2020-05-29
Replace `std::error::description` with `to_string()`


andreeaflorescu:
In newer Rust versions `description` is deprecated (since 1.42.0). We can use to_string instead.

This function seems to be heavily used in implementing display for errors.

alexandruag:
We ended up implementing the `Display` trait, which is used to automatically implement `ToString` when present..

PR #42, opened 2020-07-13
cmdline: implement std::error::Error for Error

jiangliu:
All Error enumerations except cmdline::Error have implemented
std::error::Error, so add the missing one.

Signed-off-by: Liu Jiang <gerry@linux.alibaba.com>

Please refer to #35 for more information.

### DIFF
--- a/src/cmdline/mod.rs
+++ b/src/cmdline/mod.rs
@@ -39,6 +39,8 @@ impl fmt::Display for Error {
     }
 }
 
+impl std::error::Error for Error {}
+
 /// Specialized [`Result`] type for command line operations.
 ///
 /// [`Result`]: https://doc.rust-lang.org/std/result/enum.Result.html
--- a/src/loader/x86_64/elf/mod.rs
+++ b/src/loader/x86_64/elf/mod.rs
@@ -12,8 +12,7 @@
 
 #![cfg(all(feature = "elf", any(target_arch = "x86", target_arch = "x86_64")))]
 
-use std::error::{self, Error as StdError};
-use std::fmt::{self, Display};
+use std::fmt;
 use std::io::{Read, Seek, SeekFrom};
 use std::mem;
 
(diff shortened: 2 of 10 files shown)


rbradford:
GitHub is telling me that the commit in this PR has no content.

jiangliu:
> GitHub is telling me that the commit in this PR has no content.
My bad, fixed:)

src/cmdline/mod.rs
```diff
@@ -39,6 +40,8 @@ impl fmt::Display for Error {
     }
 }
 
+impl StdError for Error {}
```
rbradford, line 43: Is this really the conventional way to do it rather than:

`impl std::error::Error for Error {}`

jiangliu: Just to keep same style with other implementations in linux-loader. Actually I prefer:
impl std::error::Error for Error {}

rbradford: If the rest of the code does it the you should continue to be consistent.

andreeaflorescu: I also prefer impl std::error::Error; It is more clear that StdError :(

src/loader/x86_64/elf/mod.rs
```diff
@@ -86,6 +86,12 @@ impl error::Error for Error {
     }
 }
 
+impl Display for Error {
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        write!(f, "Kernel Loader Error: {}", self.description())
```
andreeaflorescu, line 91: `description` is deprecated in newer rust versions. See #35

src/cmdline/mod.rs
jiangliu: let's do it:)

src/loader/x86_64/elf/mod.rs
jiangliu: Yes, both description() and cause() are deprecated in favor of Display/source().
I have just moved the position of the code without any changes:)

andreeaflorescu: Ok.

andreeaflorescu:
@jiangliu it looks like you are now fixing #35. Can you please amend you commit and add to it:
```Fixes: https://github.com/rust-vmm/linux-loader/issues/35```

jiangliu:
> @jiangliu it looks like you are now fixing #35. Can you please amend you commit and add to it:
> `Fixes: https://github.com/rust-vmm/linux-loader/issues/35`

Done

PR #43, opened 2020-07-14
[RFC] Benchmark tests & integration test demo

aghecenco:
This PR serves several purposes:
1. (the obvious one) add sample benchmark tests that use the [`criterion`](https://github.com/bheisler/criterion.rs) plugin. 2 such tests exercise PVH related code (random choice). The tests can be run locally with `cargo bench`. Note that the numbers are prone to vary significantly between hosts.
1. showcase a proposal for a starting point in the effort of adding performance tests to all `rust-vmm` crates that need them (in the future, all of them) with [`rust-vmm-ci`](https://github.com/rust-vmm/rust-vmm-ci) and dedicated Buildkite pipelines.

**Known caveat**: because the tip of the upstream `linux-loader` does not yet contain benchmarks, the test that tries to extract them & compare is bound to fail. The intention for this PR is to showcase the whole harness (such as it is in its incipient day), get feedback on the approach, then (in case of positive feedback) **first** merge the benchmark tests, then continue with the `rust-vmm-ci` and container changes.
***
The changes can be broken down as follows:
* this crate: benchmark tests are added. They are configured so as not to run as part of `cargo test`.
* [`rust-vmm-ci`](https://github.com/rust-vmm/rust-vmm-ci): a new integration test is added. To showcase how it works, this PR contains a commit that temporarily points the `rust-vmm-ci` submodule to [this branch in my `rust-vmm-ci` fork](https://github.com/aghecenco/rust-vmm-ci/tree/critcmp). The test runs `cargo bench` on the tip of the upstream crate and on the `HEAD` of the PR and compares the results with [`critcmp`](https://github.com/BurntSushi/critcmp). For the moment, the test only prints the results.
* [`rust-vmm-container`](https://github.com/rust-vmm/rust-vmm-container/): a new dev version of the container, `rustvmm/dev:v5_bench`, uploaded to Dockerhub, contains `git` and `critcmp` so that the integration test mentioned above can run.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,3 +1,4 @@
 [submodule "rust-vmm-ci"]
 	path = rust-vmm-ci
-	url = https://github.com/rust-vmm/rust-vmm-ci.git
+    url = https://github.com/aghecenco/rust-vmm-ci.git
+	branch = critcmp
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -4,15 +4,28 @@ version = "0.1.0"
 authors = ["Cathy Zhang <cathy.zhang@intel.com>"]
 edition = "2018"
 license = "Apache-2.0 AND BSD-3-Clause"
+autobenches = false
 
 [features]
 default = ["elf", "pe"]
-elf = []
 bzimage = []
+elf = []
 pe = []
 
 [dependencies]
 vm-memory = ">=0.2.0"
 
 [dev-dependencies]
+criterion = "=0.3.0"
 vm-memory = {features = ["backend-mmap"]}
+
+[[bench]]
+name = "main"
+harness = false
+
+[lib]
+bench = false # https://bheisler.github.io/criterion.rs/book/faq.html#cargo-bench-gives-unrecognized-option-errors-for-valid-command-line-options
+
+[profile.bench]
+lto = true
+codegen-units = 1
--- a/benches/main.rs
+++ b/benches/main.rs
@@ -1,0 +1,90 @@
+// Copyright 2020 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+//
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE-BSD-3-Clause file.
+//
+// SPDX-License-Identifier: Apache-2.0 AND BSD-3-Clause
+extern crate criterion;
+extern crate linux_loader;
+extern crate vm_memory;
+
+use criterion::{black_box, criterion_group, criterion_main, Criterion};
+
+#[cfg(any(target_arch = "x86", target_arch = "x86_64"))]
+mod x86_64;
+#[cfg(any(target_arch = "x86", target_arch = "x86_64"))]
+use x86_64::*;
+
+#[cfg(target_arch = "aarch64")]
+mod aarch64;
+#[cfg(target_arch = "aarch64")]
+use aarch64::*;
+
+#[cfg(target_arch = "aarch64")]
+use linux_loader::configurator::fdt::FdtBootConfigurator;
+#[cfg(any(target_arch = "x86", target_arch = "x86_64"))]
+use linux_loader::configurator::pvh::PvhBootConfigurator;
+use linux_loader::configurator::BootConfigurator;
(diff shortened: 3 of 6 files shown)


Cargo.toml
```diff
@@ -15,4 +16,12 @@ pe = []
 vm-memory = ">=0.2.0"
 
 [dev-dependencies]
+criterion = "=0.3.0"
 vm-memory = {features = ["backend-mmap"]}
+
```
alexandruag, line 21: I think it's useful to enable more optimizations than the default for the `[bench]` profile, so measurements are closer to what ppl do (or should do :D) for production builds. Seems like a good idea to enable `lto` at the very least, and potentially set `codegen-units` to `1` as well, i.e.:

```rust
[profile.bench]
lto = true
codegen-units = 1
```

.gitmodules
```diff
@@ -1,3 +1,4 @@
 [submodule "rust-vmm-ci"]
 	path = rust-vmm-ci
-	url = https://github.com/rust-vmm/rust-vmm-ci.git
+    url = https://github.com/aghecenco/rust-vmm-ci.git
```
alexandruag, line 3: Do you think it makes sense to create a RFC PR for `rust-vmm-ci` as well with the contents of that branch, so we can add further comments there?

Cargo.toml
```diff
@@ -15,4 +16,12 @@ pe = []
 vm-memory = ">=0.2.0"
 
 [dev-dependencies]
+criterion = "=0.3.0"
```
andreeaflorescu, line 19: Question, is it possible to gate this dependency on a feature? When we started using criterion in Firecracker it pretty much double the debug build time and I was wondering if we can add this dependency only when we want to run performance tests.

benches/main.rs
```diff
@@ -0,0 +1,101 @@
...
+const XEN_HVM_START_MAGIC_VALUE: u32 = 0x336ec578;
+
+fn create_guest_memory() -> GuestMemoryMmap {
+    GuestMemoryMmap::from_ranges(&[(GuestAddress(0x0), MEM_SIZE)]).unwrap()
+}
+
+fn create_elf_pvh_image() -> Vec<u8> {
+    include_bytes!(concat!(
+        env!("CARGO_MANIFEST_DIR"),
+        "/src/loader/x86_64/elf/test_elfnote.bin"
```
andreeaflorescu, line 32: Would it make sense to also test with a closer to real-life image? I was thinking we might be able to reuse the image that we are downloading in the buildkite hooks.

Also, I don't know if you were already planning this, but would it make sense to test different scenarios: loading bzimage, with/without pvh, others that might come up?

aghecenco: Of course. I intend to benchmark all the publicly exposed functionalities, including the bzImage load. I only started with 2 tests to prove the approach on the infrastructure (need *something* to run)

.gitmodules
aghecenco: * `rust-vmm-ci`: https://github.com/rust-vmm/rust-vmm-ci/pull/25
* `rust-vmm-container`: https://github.com/rust-vmm/rust-vmm-container/pull/28

aghecenco:
> it seems the `bench` CI test fails for different reasons on `x86` vs `aarch`.

Should be fixed now. The benchmarks I had initially added for demo purposes were x86-specific.

Cargo.toml
aghecenco: It doesn't look like `cargo build` takes the `criterion` dev-dependency into account:
```
root@be984d61cd88:/linux-loader# cargo clean && cargo build  --all-features
warning: dependency (vm-memory) specified without providing a local path, Git repository, or version to use. This will be considered an error in future versions
   Compiling libc v0.2.67
   Compiling vm-memory v0.2.1
   Compiling linux-loader v0.1.0 (/linux-loader)
    Finished dev [unoptimized + debuginfo] target(s) in 3.23s
```

aghecenco:
Please see https://github.com/rust-vmm/linux-loader/pull/44 for **just** the benchmarks. #44 needs to be merged first, before other the infrastructure-related PRs in `rust-vmm-container` and `rust-vmm-ci`.

aghecenco:
Closing in favor of #44

PR #44, opened 2020-07-17
Add benchmark tests for publicly exposed functions

aghecenco:
The first commit is a subset of #43, with an additional commit adding benchmarks for the bzImage loading functionality (and showcasing how to benchmark feature-gated functions).

This PR is part of a larger story of adding performance tests to rust-vmm crates and running them automatically. See https://github.com/rust-vmm/rust-vmm-ci/issues/24.

To automatically run these tests in the `linux-loader` crate, the following PRs need to be merged in this order:
1. This PR
1. https://github.com/rust-vmm/rust-vmm-container/pull/28
1. https://github.com/rust-vmm/rust-vmm-ci/pull/25
1. TODO: PR in `linux-loader` that replicates https://github.com/rust-vmm/linux-loader/pull/43/commits/6ceda5b62c8fefa726eb285a2d8df0cee075b0a3 - custom Buildkite pipeline that uses the new container and runs the new integration test

### DIFF
--- a/.buildkite/bench_pipeline.yml
+++ b/.buildkite/bench_pipeline.yml
@@ -1,0 +1,28 @@
+# Copyright 2020 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+# SPDX-License-Identifier: Apache-2.0
+steps:
+  - label: "bench-x86_64"
+    commands:
+      - pytest rust-vmm-ci/integration_tests/test_benchmark.py -s
+    retry:
+      automatic: false
+    agents:
+      platform: x86_64.metal
+    plugins:
+      - docker#v3.0.1:
+          image: "rustvmm/dev:v6"
+          always-pull: true
+          propagate-environment: true
+
+  - label: "bench-aarch64"
+    commands:
+      - pytest rust-vmm-ci/integration_tests/test_benchmark.py -s
+    retry:
+      automatic: false
+    agents:
+      platform: arm.metal
+    plugins:
+      - docker#v3.0.1:
+          image: "rustvmm/dev:v6"
+          always-pull: true
+          propagate-environment: true
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -4,15 +4,28 @@ version = "0.1.0"
 authors = ["Cathy Zhang <cathy.zhang@intel.com>"]
 edition = "2018"
 license = "Apache-2.0 AND BSD-3-Clause"
+autobenches = false
 
 [features]
 default = ["elf", "pe"]
-elf = []
 bzimage = []
+elf = []
 pe = []
 
 [dependencies]
 vm-memory = ">=0.2.0"
 
 [dev-dependencies]
+criterion = ">=0.3.0"
 vm-memory = {features = ["backend-mmap"]}
+
+[[bench]]
+name = "main"
+harness = false
+
+[lib]
+bench = false # https://bheisler.github.io/criterion.rs/book/faq.html#cargo-bench-gives-unrecognized-option-errors-for-valid-command-line-options
+
+[profile.bench]
+lto = true
+codegen-units = 1
--- a/benches/aarch64/mod.rs
+++ b/benches/aarch64/mod.rs
@@ -38,7 +38,7 @@ pub fn criterion_benchmark(c: &mut Criterion) {
     c.bench_function("configure_fdt", |b| {
         b.iter(|| {
             black_box(FdtBootConfigurator::write_bootparams::<GuestMemoryMmap>(
-                fdt_boot_params.clone(),
+                &fdt_boot_params,
                 &guest_mem,
             ))
             .unwrap();
--- a/benches/aarch64/mod.rs
+++ b/benches/aarch64/mod.rs
@@ -1,0 +1,47 @@
+// Copyright 2020 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+//
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE-BSD-3-Clause file.
+//
+// SPDX-License-Identifier: Apache-2.0 AND BSD-3-Clause
+extern crate criterion;
+extern crate linux_loader;
+extern crate vm_memory;
+
+use linux_loader::configurator::fdt::FdtBootConfigurator;
+use linux_loader::configurator::{BootConfigurator, BootParams};
+use vm_memory::{ByteValued, GuestAddress, GuestMemoryMmap};
+
+use criterion::{black_box, Criterion};
+
+const MEM_SIZE: usize = 0x100_0000;
+const FDT_MAX_SIZE: usize = 0x20;
+
+fn create_guest_memory() -> GuestMemoryMmap {
+    GuestMemoryMmap::from_ranges(&[(GuestAddress(0x0), MEM_SIZE)]).unwrap()
+}
+
+#[derive(Clone, Copy, Default)]
+pub struct FdtPlaceholder([u8; FDT_MAX_SIZE]);
+
+unsafe impl ByteValued for FdtPlaceholder {}
+
+fn build_fdt_boot_params() -> BootParams {
+    let fdt = FdtPlaceholder([0u8; FDT_MAX_SIZE]);
+    let fdt_addr = GuestAddress((MEM_SIZE - FDT_MAX_SIZE - 1) as u64);
+    BootParams::new::<FdtPlaceholder>(&fdt, fdt_addr)
+}
+
+pub fn criterion_benchmark(c: &mut Criterion) {
+    let guest_mem = create_guest_memory();
+    let fdt_boot_params = build_fdt_boot_params();
+    c.bench_function("configure_fdt", |b| {
+        b.iter(|| {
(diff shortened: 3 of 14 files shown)


benches/main.rs
```diff
@@ -64,6 +67,23 @@ pub fn criterion_benchmark(c: &mut Criterion) {
     });
 }
 
+#[cfg(all(feature = "bzimage", any(target_arch = "x86", target_arch = "x86_64")))]
+pub fn criterion_benchmark_bzimage(c: &mut Criterion) {
+    let guest_mem = create_guest_memory();
+    let image = create_bzimage();
```
alexandruag, line 73: Can we add this `let` and the following `c.bench_function` call to the `x86` specific `criterion_benchmark` (if the previous suggestion about that makes sense), within a block guarded by a  `#[cfg(feature = "bzimage")]`, so we don't have to define the `criterion_benchmark_bzimage` function altogether (and the `criterion_group` variations)?

aghecenco: Nope. The `bzimage` benchmark needs to be in another group because it's run for fewer iterations. The number of iterations is defined per group. I expect such cases, where different benchmarks need various run configurations, to show up more often as we add more and more. I couldn't think of any other way of combining this with feature-gating, other than the `nop` bench. Any ideas?

benches/aarch64/mod.rs
```diff
@@ -0,0 +1,30 @@
...
+extern crate linux_loader;
+extern crate vm_memory;
+
+use linux_loader::configurator::BootParams;
+use vm_memory::{ByteValued, GuestAddress, GuestMemoryMmap};
+
+const MEM_SIZE: usize = 0x100_0000;
+const FDT_MAX_SIZE: usize = 0x20;
+
+pub fn create_guest_memory() -> GuestMemoryMmap {
```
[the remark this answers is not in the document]
aghecenco: I think we could add an arch-independent `utils` module, it just seemed overkill to do so for a one-liner function (though I think it will become necessary with the next arch-independent utility). Would you prefer that I add it now or leave it for later?

benches/main.rs
alexandruag: Oh I see. In that case, can we rename `criterion_benchmark_nop` to `criterion_benchmark_bzimage`, and put a `cfg` gated call to `x86_64::benchmark_bzimage` inside? It looks like that way we don't need the additional `#[cfg(not(feature = "bzimage"))] criterion_group! { ...` block, since the function effectively turns into nop when the feature is not enabled. Don't have any other ideas :(

benches/aarch64/mod.rs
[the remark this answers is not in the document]
alexandruag: Would be tidier to add it right now, but it's by no means a blocker :D

benches/main.rs
aghecenco: I like the separate `nop` better because it can stay unchanged for as many features as we gate-bench. It runs nothing for the minimum number of iterations (10). If we get rid of it, we'll run the same nothing but for more iterations when the feature is off (in this case, 100). But, given that we're only doing this for *one* feature, I don't feel strongly for it. How do you see feature-gating benchmarks with more than 1 feature?

alexandruag: I think the number of iterations for nothing doesn't matter that much (it doesn't even show up in the results), but yeah it was just an alternative and the current one is perfectly fine as well.

aghecenco:
CI is failing because the custom hook that downloads the `bzimage` used in the unit & perf tests is failing. See https://buildkite.com/rust-vmm/linux-loader-ci/builds/225#f5550401-48e8-4c51-a7ba-ec464740f49d for full output.

TLDR the .deb that the hook downloads is no longer available at that address. Will update the hook in a separate PR as the CI remains blocked otherwise.

aghecenco:
Blocked by #45

Cargo.toml
```diff
@@ -4,15 +4,28 @@ version = "0.1.0"
...
-elf = []
 bzimage = []
+elf = []
 pe = []
 
 [dependencies]
 vm-memory = ">=0.2.0"
 
 [dev-dependencies]
+criterion = "=0.3.0"
```
andreeaflorescu, line 19: Any reason to block the criterion version to 0.3.0? If yes, can you also add a comment about this so that we can later remember it?

benches/aarch64/mod.rs
```diff
@@ -0,0 +1,48 @@
...
+// found in the LICENSE-BSD-3-Clause file.
+//
+// SPDX-License-Identifier: Apache-2.0 AND BSD-3-Clause
+extern crate criterion;
+extern crate linux_loader;
+extern crate vm_memory;
+
+use linux_loader::configurator::fdt::FdtBootConfigurator;
+use linux_loader::configurator::BootConfigurator;
+use linux_loader::configurator::BootParams;
```
andreeaflorescu, line 13: nit: these can be imported on the same line.

benches/aarch64/mod.rs
```diff
@@ -0,0 +1,48 @@
...
+    BootParams::new::<FdtPlaceholder>(&fdt, fdt_addr)
+}
+
+pub fn criterion_benchmark(c: &mut Criterion) {
+    let guest_mem = create_guest_memory();
+    let fdt_boot_params = build_fdt_boot_params();
+    c.bench_function("configure_fdt", |b| {
+        b.iter(|| {
+            black_box(FdtBootConfigurator::write_bootparams::<GuestMemoryMmap>(
+                fdt_boot_params.clone(),
```
andreeaflorescu, line 42: With this approach it looks like we're also including the time it takes to clone the fdt_boot_params into consideration. Do we have information of how expensive clone might be in this scenario? I am wondering if we can re-write it somehow such that clone is not executed in the benchmark function.

aghecenco: Indeed, other artefacts made their way into the bench loop (the `clone`s, a `Cursor::new`). I couldn't find a way to avoid them. I'll check how long they take.

.buildkite/bench_pipeline.yml
```diff
@@ -0,0 +1,14 @@
+# Copyright 2020 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+# SPDX-License-Identifier: Apache-2.0
+steps:
```
andreeaflorescu, line 3: Looks like this can be now updated.

Cargo.toml
aghecenco: Updated

benches/aarch64/mod.rs
aghecenco: Done

.buildkite/bench_pipeline.yml
aghecenco: Not yet. If I update the pipeline with this PR, the CI will try to run the benchmarks on upstream/master, where they're obviously not merged yet => CI will fail. This PR needs to go in first, then the pipeline will be updated in a new one.

benches/aarch64/mod.rs
aghecenco: Checked and:
* `Cursor::new` takes a meager ~1% of the time it takes to `load` the kernel
* but the `clone` takes 50% of the total `write_bootparams` time

I can't think of a way to eliminate the `clone` other than refactoring `write_bootparams` to take a `BootParams&` instead, if that's possible. Thoughts?

aghecenco: Wasn't that hard to pass the `BootParams` by ref, and the benchmark numbers are looking better :+1: Nice catch!

.buildkite/bench_pipeline.yml
andreeaflorescu: This is no longer the case. I've updated the test_bench such that when master has no performance tests, it just runs the test on the PR. PR for reference: https://github.com/rust-vmm/rust-vmm-ci/pull/31. We modified this behaviour while you were on vacation :)) thought it's a bug.

benches/aarch64/mod.rs
```diff
@@ -0,0 +1,47 @@
+// Copyright 2020 Amazon.com, Inc. or its affiliates. All Rights Reserved.
```
andreeaflorescu, line 1: Just realized, we also need to exclude the `benches` directory when computing the coverage. Can you also update the coverage_config file?

aghecenco:
The 1.6% coverage drop is not caused by `kcov` counting the sources in `benches/`. It's somehow related to the migration to `rust-vmm-container` v6, which brings in a hefty `rustc` update (1.39.0 -> 1.44.1).

Container | Instrumented lines | Executed lines
-- | -- | --
v5 | 453 | 359
v6 | 442 | 350

benches/aarch64/mod.rs
aghecenco: Done, the coverage test actually fails with the newer `rustc`, so I had to decrease the expected value :(

andreeaflorescu:
> The 1.6% coverage drop is not caused by `kcov` counting the sources in `benches/`. It's somehow related to the migration to `rust-vmm-container` v6, which brings in a hefty `rustc` update (1.39.0 -> 1.44.1).
> Container 	Instrumented lines 	Executed lines
> v5 	453 	359
> v6 	442 	350

Had the same problem in event-manager. I don't know why I forgot about it :laughing:  sorry for making you go off track.

PR #45, opened 2020-07-23
buildkite: update bzimage location in hook

aghecenco:
Debian changed the structure of their ftp mirror, so we need to download the bzimage used in the tests from the new location.


aghecenco:
Too tightly coupled with #44, needs the nop pipeline introduced there to pass CI. Closing, will cherry pick on top of #44 instead

PR #46, opened 2020-07-24
Enable performance tests at every PR

aghecenco:
**This PR is on top of #44. Please review only the last 3 commits. #44 should be merged first.**

This PR hooks `linux-loader` to the recently introduced bench integration test in rust-vmm-ci. The benchmarks introduced by #44 will be run at every PR and the test will print comparison results relative to the upstream code.

### DIFF
--- a/.buildkite/bench_pipeline.yml
+++ b/.buildkite/bench_pipeline.yml
@@ -1,0 +1,14 @@
+# Copyright 2020 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+# SPDX-License-Identifier: Apache-2.0
+steps:
+  - label: "nop"
+    commands:
+     - true
+    retry:
+      automatic: false
+    agents:
+      platform: x86_64.metal
+    plugins:
+      - docker#v3.0.1:
+          image: "rustvmm/dev:v5"
+          always-pull: true
--- a/.buildkite/bench_pipeline.yml
+++ b/.buildkite/bench_pipeline.yml
@@ -1,14 +1,26 @@
 # Copyright 2020 Amazon.com, Inc. or its affiliates. All Rights Reserved.
 # SPDX-License-Identifier: Apache-2.0
 steps:
-  - label: "nop"
+  - label: "bench-x86_64"
     commands:
-     - true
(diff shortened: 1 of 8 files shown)


.buildkite/pipeline.yml
```diff
@@ -10,7 +10,7 @@ steps:
       platform: x86_64.metal
     plugins:
       - docker#v3.0.1:
-          image: "rustvmm/dev:v5"
+          image: "rustvmm/dev:v6"
```
alexandruag, line 13: Does it make sense/is it possible to have a variable that defines the desired version of the container so we don't have to explicitly mention it in multiple places? (maybe we can even share it among multiple files)? I think I saw something like this somewhere, but I might be wrong :-s

aghecenco: You saw it in a comment and a bunch of readmes. The closest thing I found to applying this to the pipeline is regenerating the pipeline at every run: https://buildkite.com/docs/pipelines/defining-steps#dynamic-pipelines
I can't figure out another way of substituting variables in the yaml.

alexandruag: Sure, maybe there's some buildkite-specific or other feature, but we can look out for that in a separate setting. Looks good, it seems we just have fix some things the CI's complaining about.

.buildkite/bench_pipeline.yml
```diff
@@ -0,0 +1,26 @@
...
+    commands:
+      - pytest rust-vmm-ci/integration_tests/test_benchmark.py -s
+    retry:
+      automatic: false
+    agents:
+      platform: x86_64.metal
+    plugins:
+      - docker#v3.0.1:
+        image: "rustvmm/dev:v6"
+        always-pull: true
```
andreeaflorescu, line 14: These lines have a whitespace problem; they need to be indented.

This is what triggers the following error in the Buildkite pipeline:
```
2020-07-24 11:44:30 WARN   POST https://agent.buildkite.com/v3/jobs/033d13f0-9ad6-4863-b9b1-e6411cb53262/pipelines: 422 `image` is invalid: configuration must be defined with a hash (Attempt 1/60 Retrying in 5s)
```

aghecenco:
Closing in favor of #44

PR #53, opened 2020-11-27
Specify serial console & virtio MMIO devices to the kernel command line

aghecenco:
This PR extends the `Cmdline` with the possibility of specifying serial console and virtio MMIO device configurations, which are transparently stringified and appended to the command line.

### DIFF
--- a/src/cmdline/mod.rs
+++ b/src/cmdline/mod.rs
@@ -11,31 +11,42 @@
 use std::fmt;
 use std::result;
 
+use vm_memory::{Address, GuestAddress, GuestUsize};
+
 /// The error type for command line building operations.
 #[derive(Debug, PartialEq)]
 pub enum Error {
     /// Operation would have resulted in a non-printable ASCII character.
     InvalidAscii,
+    /// Invalid device passed to the kernel command line builder.
+    InvalidDevice(String),
     /// Key/Value Operation would have had a space in it.
     HasSpace,
     /// Key/Value Operation would have had an equals sign in it.
     HasEquals,
+    /// 0-sized virtio MMIO device passed to the kernel command line builder.
+    MmioSize,
     /// Operation would have made the command line too large.
     TooLarge,
 }
 
 impl fmt::Display for Error {
     fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
-        write!(
-            f,
-            "{}",
-            match *self {
-                Error::InvalidAscii => "String contains a non-printable ASCII character.",
-                Error::HasSpace => "String contains a space.",
-                Error::HasEquals => "String contains an equals sign.",
-                Error::TooLarge => "Inserting string would make command line too long.",
-            }
-        )
+        match *self {
+            Error::InvalidAscii => write!(f, "String contains a non-printable ASCII character."),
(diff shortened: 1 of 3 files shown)


src/cmdline/mod.rs
```diff
@@ -209,6 +220,114 @@ impl Cmdline {
...
+    /// # use linux_loader::cmdline::*;
+    /// # use std::ffi::CString;
+    /// let mut cl = Cmdline::new(50);
+    /// cl.add_serial_console("ttyS0", vec!["9600"]).unwrap();
+    /// let cl_cstring = CString::new(cl).unwrap();
+    /// assert_eq!(cl_cstring.to_str().unwrap(), "console=ttyS0,9600");
+    /// ```
+    ///
+    /// [documentation]: https://www.kernel.org/doc/html/v4.16/admin-guide/kernel-parameters.html
+    pub fn add_serial_console<T: AsRef<str>>(&mut self, device: T, options: Vec<T>) -> Result<()> {
```
rbradford, line 249: This should be more generally `add_console` as we can use it to specify `hvc0` which is the virtio-console.

andreeaflorescu:
LGTM, besides the suggestion the @alexandruag made :+1:

src/cmdline/mod.rs
```diff
@@ -209,6 +220,114 @@ impl Cmdline {
...
+        if size % GB_MULT == 0 {
+            return format!("{}G", size / GB_MULT);
+        }
+        if size % MB_MULT == 0 {
+            return format!("{}M", size / MB_MULT);
+        }
+        if size % KB_MULT == 0 {
+            return format!("{}K", size / KB_MULT);
+        }
+        size.to_string()
```
andreeaflorescu, line 329: Is the size valid if none of the standard suffixes are appropriate? I was thinking if we should return an error in case none of them matches.

aghecenco: It's valid. The kernel trims out the suffix, if present, and parses the number normally, so it's valid even if expressed in bytes. No further validations appear to be made.
Parsing the device: https://elixir.bootlin.com/linux/latest/source/drivers/virtio/virtio_mmio.c#L636
Parser: https://elixir.bootlin.com/linux/latest/source/lib/cmdline.c#L125

src/cmdline/mod.rs
```diff
@@ -209,6 +212,45 @@ impl Cmdline {
...
+    /// # use linux_loader::cmdline::*;
+    /// # use std::ffi::CString;
+    /// let mut cl = Cmdline::new(50);
+    /// cl.add_console("ttyS0", vec!["9600"]).unwrap();
+    /// let cl_cstring = CString::new(cl).unwrap();
+    /// assert_eq!(cl_cstring.to_str().unwrap(), "console=ttyS0,9600");
+    /// ```
+    ///
+    /// [documentation]: https://www.kernel.org/doc/html/latest/admin-guide/kernel-parameters.html
+    pub fn add_console<T: AsRef<str>>(&mut self, device: T, options: Vec<T>) -> Result<()> {
```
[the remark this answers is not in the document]
aghecenco: Done

andreeaflorescu (approved):

alexandruag (approved):

PR #55, opened 2020-12-30
Run unit tests on local dev machines

andreeaflorescu:
This PR removes the buildkite hooks and makes it possible to run the unit tests on local dev machines as well by downloading the required resources for running the test as part of a rust build script.


alexandruag (approved):

jiangliu (approved):

rbradford:
This breaks the compilation of Cloud Hypervisor on Fedora/RedHat machines. Please consider an alternative which doesn't require using dpkg-deb in the build.rs step.

andreeaflorescu:
@rbradford is it possible to run the rust-vmm docker container on those machines?

PR #56, opened 2021-01-05
Partially revert "add build.rs that dlds the test resources"

rbradford:
This reverts commit b18d3878892ecbef70694ba6b72efd4e7a0c3c47. With the
exception of changing from curl to wget

Making every user of linux-loader run the scripts required to download
the kernel and extract it is a problem as it adds extra package
requirements on projects that depend on it as well as requiring a
network connection.


andreeaflorescu:
#57 is now merged. We can close this.

PR #57, opened 2021-01-05
remove build.rs

andreeaflorescu:
Download the needed resources before running the tests.
Do not use include_bytes! macro because it makes cargo check fail when the bzimage path does not work.

Build.rs is not a feasible solution because the build of linux-loader will fail on machines without network connectivity.

Reported-by: @rbradford


rbradford (dismissed):
This addresses the issue I was seeing with build.rs in projects that use linux-loader as a dependency.

rbradford (approved):

alexandruag (approved):

Issue #66, opened 2021-02-25
Update `align_up` function to return an u64 instead of usize


andreeaflorescu:
Update `align_up` function to return an u64 instead of usize. We also need to check all the usages of `align_up` and validate that they are lossless.

andreeaflorescu:
Fixed by: https://github.com/rust-vmm/linux-loader/pull/70

PR #70, opened 2021-03-08
elf: `align_up` usage susceptible to data loss

iulianbarbu:
In response to #66.

* While using `aling_up` we're converting the function result
to other data types. The conversions were double checked and
comments were added to clarify why they are safe.

### DIFF
--- a/src/loader/x86_64/elf/mod.rs
+++ b/src/loader/x86_64/elf/mod.rs
@@ -313,8 +313,6 @@ where
     // the PVH boot protocol.
     const XEN_ELFNOTE_PHYS32_ENTRY: u32 = 18;
 
-    let n_align = phdr.p_align;
-
     // Seek to the beginning of the note segment.
     kernel_image
         .seek(SeekFrom::Start(phdr.p_offset))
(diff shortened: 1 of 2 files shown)


src/loader/x86_64/elf/mod.rs
```diff
@@ -343,15 +343,28 @@ where
         }
 
         // Skip the note header plus the size of its fields (with alignment).
-        let namesz_aligned = align_up(u64::from(nhdr.n_namesz), n_align)?;
-        let descsz_aligned = align_up(u64::from(nhdr.n_descsz), n_align)?;
+        // How to compute next notes fields can be found in this man page:
+        // https://man7.org/linux/man-pages/man5/elf.5.html.
+        let namesz_aligned = align_up(u64::from(nhdr.n_namesz), 4)?;
```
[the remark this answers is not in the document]
iulianbarbu: I've done some research and my understanding is that `p_align` refers to segments alignment rather than `name` or `desc` notes fields alignment. There are also other resources that point to 4-byte alignment [1], [2], [3], [4]. [1] mentions about an alignment dependent on the arch reg size (4 bytes vs 8 bytes), so I think we lack this sort of logic as well. [5] is an old mail thread regarding how should notes be aligned on 64bit systems and the overall idea was that most applications (gdb, gcc and other binutils tools) and the linux kernel expect 4 bytes alignment even on 64bit systems. 

I suggest having only the alignment adjustment, letting arch dependent implementation arise as a requirement, if needed (it is unlikely, low priority at the moment and it is also dependent on the gABI implementation of the build toolchain used).

[1] https://refspecs.linuxbase.org/elf/gabi4+/ch5.pheader.html#note_section - LSB details on ELF notes
[2] http://www.sco.com/developers/devspecs/gabi41.pdf - Generic ABI, page 81
[3] http://lkml.iu.edu/hypermail/linux/kernel/0611.1/0538.html - discussion around modifying note alignment on 64 bit platforms.
[4] linux kernel elf core dumper `notesize` - https://elixir.bootlin.com/linux/latest/source/fs/binfmt_elf.c#L1429

src/loader/x86_64/elf/mod.rs
```diff
@@ -343,15 +341,26 @@ where
         }
 
         // Skip the note header plus the size of its fields (with alignment).
-        let namesz_aligned = align_up(u64::from(nhdr.n_namesz), n_align)?;
-        let descsz_aligned = align_up(u64::from(nhdr.n_descsz), n_align)?;
+        let namesz_aligned = align_up(u64::from(nhdr.n_namesz), phdr.p_align)?;
+        let descsz_aligned = align_up(u64::from(nhdr.n_descsz), phdr.p_align)?;
+
+        // `namesz` and `descsz` are both `u32`s. They refer also to 4 byte sized and aligned memory
```
andreeaflorescu, line 347: The commnet also needs to be updated as it is not 4-bytes.

iulianbarbu: Done.

andreeaflorescu (approved):

aghecenco (approved):

PR #74, opened 2021-03-17
Added reproducible steps for generating test images

andreeaflorescu:
The images we previously used in test were not easy to recreate, and their source (for most of them) was unknown.
To make sure we do not have any licensing problems, and also to make it easy to create new images, all of them were regenerated  from scratch. The documentation now includes reproduction steps for every image.

As a summary, we now have 3 types of images:
- generated by fuzzers (these are named with `fuzz_` so that are easy to identify)
- generated with ELFIO
- generated from kernel sources

This PR also includes some other tiny cleanup:
- fixed clippy errors
- removed outdated section from the testing document


docs/TESTING.md
```diff
@@ -45,3 +45,29 @@ make allnoconfig
...
+|-------------|-----------------------|
+| bad_align_writer.cpp | test_bad_align.bin |
+
+#### Example for generating `test_bad_align.bin`
+
+```bash
+git clone git@github.com:serge1/ELFIO.git
+
+# Copy the bad_align_writer.cpp file from this repo to the ELFIO path.
+cp ${LINUX_LOADER_PATH}/docs/elfio/overflow_elfnote.cpp ${ELFIO_PATH/
```
aghecenco, line 72:
```suggestion
cp ${LINUX_LOADER_PATH}/docs/elfio/overflow_elfnote.cpp ${ELFIO_PATH}/
```

I'd also enclose the paths within double quotes, particularly because you dereference variables, but that's just a style nit.

docs/TESTING.md
```diff
@@ -17,28 +17,64 @@ docker run -it \
...
-           rustvmm/dev:v${container_version}
-cd linux-loader/
-cargo test
+git clone git://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git
+git checkout v4.20
+
+# To be able to run the following command, you'll need to install
+# flex and bison.
+make allnoconfig
+make Image
```
aghecenco, line 45: I don't know if it's because of my setup, but this only worked for me on an `aarch64` machine. Can you `make Image` on x86_64?

docs/TESTING.md
```diff
@@ -17,28 +17,64 @@ docker run -it \
...
+head -c 4096 arch/arm64/boot/Image > test_image.bin
+```
+
+### Generating `ELF` images with [ELFIO](https://github.com/serge1/ELFIO)
+
+All the ELF files used in `linux-loader` are generated using the `ELFIO` tool.
```
aghecenco, line 51: I think it's worth a mention that `ELFIO` requires a compiler with builtin C++11 support. For `gcc`/`g++`, that'd be at least `g++-6`. I tried to work around it with my older compiler but it wouldn't budge :)

docs/TESTING.md
aghecenco: Also, there's no `overflow_elfnote.cpp`. Was it supposed to be `bad_align_writer.cpp`?

aghecenco: And the intermediary directory is `elfio_files`, not `elfio`.

docs/TESTING.md
```diff
@@ -17,28 +17,64 @@ docker run -it \
...
+```bash
+git clone git@github.com:serge1/ELFIO.git
+
+# Copy the bad_align_writer.cpp file from this repo to the ELFIO path.
+cp ${LINUX_LOADER_PATH}/docs/elfio/overflow_elfnote.cpp ${ELFIO_PATH/
+cd ${ELFIO_PATH}
+# The images in this repo were built from the`57e614a` commit.
+git checkout 57e614a
+
+g++ bad_align_writer.cpp -o bad_align_writer -I.
```
aghecenco, line 77: For folks stuck with an older compiler, I think it'd be helpful to add the `-std=c++11` flag in all the ELFIO g++ invocations so that it works regardless of gcc version.

docs/TESTING.md
andreeaflorescu: No, this is only for aarch64. The PE image is only used on aarch64 as well.

docs/TESTING.md
andreeaflorescu: Looks like this is indeed outdated. I'll update it.

docs/TESTING.md
aghecenco: Re-reading this I realize that I inadvertently said something dumb - older compilers (pre v6) **can** build C++11 (and newer) syntax as well, but you need to specify `-std=c++11` explicitly. The `g++` commands below work if you add `-std=c++11`. If you, however, try to build ELFIO itself, I can't figure out how to trick the `automake` magic into appending `-std=c++11` in `CPPFLAGS`, so older compilers can't build it. Nevertheless, turns out you don't need to build ELFIO to get the commands below to work.

andreeaflorescu: Yap, building ELFIO was kinda hard (i don't even remember what I had to install), but it shouldn't be needed for building the binaries.

docs/TESTING.md
andreeaflorescu: I explicitly added the flag you mentioned in the g++ call.

docs/TESTING.md
andreeaflorescu: I explicitly called out that these steps work on aarch64.

docs/TESTING.md
andreeaflorescu: Updated.

docs/TESTING.md
iulianbarbu: The `make Image` rule works on `x86_64` but requires cross compilation. Should we mention that these steps work out of the box only on aarch64 machines?

aghecenco: Looks like Andreea added an explicit mention: `The PE image used by the linux-loader unit tests is built on an aarch64
machine using the following commands:...`

alexandruag (approved):

PR #75, opened 2021-04-09
Add UEFI image loader for arm64.

jongwu:
UEFI image for virtual platform on arm64 is not formatted,
so I add a specific loader for it here.

Signed-off-by: Jianyong Wu <jianyong.wu@arm.com>

@aghecenco @andreeaflorescu

### DIFF
--- a/src/loader/aarch64/pe/mod.rs
+++ b/src/loader/aarch64/pe/mod.rs
@@ -50,6 +50,14 @@ pub enum Error {
     InvalidImageMagicNumber,
     /// Invalid base address alignment
     InvalidBaseAddrAlignment,
+    /// Unable to seek to UEFI image start.
+    SeekUefiStart,
+    /// Unable to seek to UEFI image end.
+    SeekUefiEnd,
+    /// UEFI image too big.
+    UefiTooBig,
+    /// Unable to read UEFI image
+    ReadUefiImage,
 }
 
 impl fmt::Display for Error {
@@ -61,9 +69,13 @@ impl fmt::Display for Error {
             Error::ReadDtbImage => "unable to read DTB image",
             Error::SeekDtbStart => "unable to seek DTB start",
             Error::SeekDtbEnd => "unable to seek DTB end",
+            Error::ReadUefiImage => "unable to read UEFI image",
+            Error::SeekUefiStart => "unable to seek UEFI start",
+            Error::SeekUefiEnd => "unable to seek UEFI end",
             Error::InvalidImage => "invalid Image",
             Error::InvalidImageMagicNumber => "invalid Image magic number",
             Error::DtbTooBig => "device tree image too big",
+            Error::UefiTooBig => "UEFI image too big",
             Error::ReadKernelImage => "unable to read kernel image",
             Error::InvalidBaseAddrAlignment => "base address not aligned to 2 MB",
         };
@@ -168,6 +180,38 @@ impl KernelLoader for PE {
     }
 }
 
+/// Load UEFI image in guest memory
+///
+/// # Arguments
+///
+/// * `guest_mem` - Guest memory address where UEFI should be loaded.
+/// * `guest_addr` - The offset in guest_mem from which the UEFI image will be loaded.
+/// * `uefi_image` - The UEFI image
+#[cfg(target_arch = "aarch64")]
+pub fn load_uefi<F, M: GuestMemory>(
+    guest_mem: &M,
+    guest_addr: GuestAddress,
+    uefi_image: &mut F,
+) -> Result<()>
+where
+    F: Read + Seek,
+{
+    let uefi_size = uefi_image
+        .seek(SeekFrom::End(0))
+        .map_err(|_| Error::SeekUefiEnd)? as usize;
+
+    // edk2 image on virtual platform does not larger than 3M
+    if uefi_size > 0x300000 {
+        return Err(Error::UefiTooBig.into());
+    }
+    uefi_image
+        .seek(SeekFrom::Start(0))
+        .map_err(|_| Error::SeekUefiStart)?;
+    guest_mem
+        .read_exact_from(guest_addr, uefi_image, uefi_size)
+        .map_err(|_| Error::ReadUefiImage.into())
+}
+
 /// Writes the device tree to the given memory slice.
 ///
 /// # Arguments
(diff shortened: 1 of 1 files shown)


src/loader/aarch64/pe/mod.rs
```diff
@@ -168,6 +180,40 @@ impl KernelLoader for PE {
     }
 }
 
+/// load uefi for aarch64
+/// uefi image for virtual machine on arm64 is not formatted, so
+/// a special loader for it is needed.
```
andreeaflorescu, line 185: I don't think this needs to be part of the public interface documentation.

src/loader/aarch64/pe/mod.rs
```diff
@@ -168,6 +180,40 @@ impl KernelLoader for PE {
     }
 }
 
+/// load uefi for aarch64
```
andreeaflorescu, line 183: nit: can you format the comment similar to how other comments are formatted: `Load UEFI image in guest memory`.

andreeaflorescu (commented):
I don't really know much about UEFI. Do you also have an example of how this is used?

Can you also add some tests?

src/loader/aarch64/pe/mod.rs
```diff
@@ -168,6 +180,40 @@ impl KernelLoader for PE {
...
+/// uefi image for virtual machine on arm64 is not formatted, so
+/// a special loader for it is needed.
+///
+/// # Arguments
+///
+/// * `guest_mem` - Guest memory address where UEFI should be loaded.
+/// * `guest_addr` - The offset in guest_mem from which the UEFI image will be loaded.
+/// * `uefi_image` - The UEFI image
+#[cfg(target_arch = "aarch64")]
+pub fn load_uefi<F, M: GuestMemory>(
```
michael2012z, line 193: The code of this functions is rather similar to `load_dtb`, can you refactor to reduce some duplicate code?

jongwu:
@andreeaflorescu -, Thanks for reply. 
Yeah, It's fine to use this UEFI image loader even having little knowledge about it. You can simply take UEFI image like a kernel image and load it. As the kernel image format is different from the UEFI image on arm64,  we need some twists on the kernel loader. Roughly, remove the header check and set the entry point of UEFI image as its very beginning then we get a new loader. That is what this patch does. So easy isn't? Here is an example of how to use it:
`let entry_point = linux_loader::loader::pe::load_uefi(`
 `               mem.deref(),                                              // Guest memory  where UEFI image should be loaded`
 `               GuestAddress(arch::get_uefi_start()),         // The offset in guest memory from which the UEFI image will be loaded`
`                &uefi_image,`
`                )`
we load the uefi_image to mem.deref() at GuestAddress(arch::get_uefi_start()) and get the entry point in "entry_point", then we can set the PC in guest as entry_point and begin to run.

src/loader/aarch64/pe/mod.rs
jongwu: ok, I will remove it.

src/loader/aarch64/pe/mod.rs
jongwu: Ok, thanks

rbradford:
For loading a UEFI/OVMF on Cloud Hypervisor with x86-64 we used the PVH (like Xen) support so that we didn't need to modify linux-loader or the VMM. I see that ArmVirtPkg also supports Xen: https://github.com/tianocore/edk2/blob/master/ArmVirtPkg/ArmVirtXen.fdf

If you can make that work then I think that would a a cleaner approach.

rbradford:
> For loading a UEFI/OVMF on Cloud Hypervisor with x86-64 we used the PVH (like Xen) support so that we didn't need to modify linux-loader or the VMM. I see that ArmVirtPkg also supports Xen: https://github.com/tianocore/edk2/blob/master/ArmVirtPkg/ArmVirtXen.fdf
> 
> If you can make that work then I think that would a a cleaner approach.

Oh, I think i misunderstood. This is a loader for the UEFI PE format rather than for a UEFI firmware blob. What's your use case for this?

jongwu:
Hello @rbradford -, for now, the UEFI image on arm64 virtual platform (eg, qemu, kvmtool) is not formatted, not a pe or elf. Unfortunately, we have no such a loader to load a file without header in it. 
In my case, I want to introduce UEFI support in cloud-hypervisor on arm64, but I can't load the UEFI as there is lack of a suitable loader. That's why I want to introduce a new one here.

jongwu:
Hi @andreeaflorescu -, any comments?
I have updated this patch, adding a test for load_uefi and amending the annotation.

andreeaflorescu:
> Hi @andreeaflorescu -, any comments?
> I have updated this patch, adding a test for load_uefi and amending the annotation.

Thanks, I can take another look in a couple of days. Can you share an example of how this is used in practice? Do you happen to have a patch that adds this support in Cloud Hypervisor?

@rbradford since it looks like this support is going to end up in Cloud Hypervisor, can you also please take a look at this PR?

michael2012z:
@andreeaflorescu , there is a PR in Cloud-hypervisor that includes the WIP commits for UEFI & ACPI support for AArch64. 

The code to call this new function is like this: https://github.com/cloud-hypervisor/cloud-hypervisor/pull/2241/commits/9742ecd60346177f8ae48736785eec527227cb58#diff-9c91c6326e9d86ef107a59aed3f13a9bb91790f9f4be1668d22cda12a28bc28cR905

The UEFI binary is header-less, it's impossible to recognize it by format. 
So the logic is: At first try to load the binary as PE; if the format is wrong, try to load it as UEFI.

jongwu:
Hello @andreeaflorescu -, could you please take a look of this patch? We have a [PR](https://github.com/cloud-hypervisor/cloud-hypervisor/pull/2680) in Cloud Hypervisor to enable Acpi on arm64 which depends on this patch.  Also,  you can find the use case of the new handler added in this patch [here](https://github.com/cloud-hypervisor/cloud-hypervisor/pull/2680)

We have blocked here so long, thus I'm very appreciate if there is any progress on this PR. Thanks!

jongwu:
Hello @andreeaflorescu -, could you please take a look of this patch? We have a [PR](https://github.com/cloud-hypervisor/cloud-hypervisor/pull/2680) in Cloud Hypervisor to enable Acpi on arm64 which depends on this patch.  Also,  you can find the use case of the new handler added in this patch [here](https://github.com/cloud-hypervisor/cloud-hypervisor/pull/2680)

We have blocked here so long, thus I'm very appreciate if there is any progress on this PR. Thanks!

jongwu:
Hello @andreeaflorescu -, could you please take a look of this patch? We have a [PR](https://github.com/cloud-hypervisor/cloud-hypervisor/pull/2680) in Cloud Hypervisor to enable Acpi on arm64 which depends on this patch.  Also,  you can find the use case of the new handler added in this patch [here](https://github.com/cloud-hypervisor/cloud-hypervisor/pull/2680)

We have blocked here so long, thus I'm very appreciate if there is any progress on this PR. Thanks!

andreeaflorescu:
@jongwu I can only provide high level feedback, and that might not be very meaningful. Can someone from Cloud Hypervisor review this since it is going to be used there? You should not block on my approval as I really do not have enough context.

rbradford (commented):
It looks like it doesn't do much more than read the file into guest memory. Maybe it makes sense to just do that in CH?

jongwu:
Thanks @rbradford @andreeaflorescu -, since we can do this change in CLH, I will close this.

PR #94, opened 2021-09-10
cmdline: fix bug in cmdline handling

### LINKED ISSUE #92: cmdline: port fix for issue reported in Firecracker
An [issue](https://github.com/firecracker-microvm/firecracker/issues/2709) was opened against FIrecracker regarding the command line functionality. 
We are working on a [fix](https://github.com/firecracker-microvm/firecracker/pull/2716) which should be brought over to rust-vmm's linux-loader crate too.

dianpopa:
Fixes #92.
As per the documentation, everything that is
passed after "--" to the commandline is considered
an argument to init.
However, in the current implementation we are dynamically
adjusting the commandline by appending system specific information
(like mmio devices, root device).
As such, the region after "--" would contain more than init args.

Signed-off-by: Diana Popa <dpopa@amazon.com>


andreeaflorescu:
I misunderstood the problem initially, but I looked at the bug report (https://github.com/firecracker-microvm/firecracker/issues/2709) and things are a bit more clear now.

From the kernel command line point of view, `boot_args` does not have any meaning, so I do not think we should add it here. I initially thought that you want to add a function for specifying the `init` arguments.

I would propose we fix this problem using the following approach:
- In Firecracker, split the `boot_args` by `--`. This will give us the regular kernel parameters, and the init parameters.
- In the `cmdline` utility, create 2 regions: one for regular parameters, and one for init parameters. This would follow the kernel command line specification, so it should be pretty straightforward to use. The `insert_init_str` should behave the same way as `insert_str`, just that it inserts the string in the `init` region instead of at the beginning. The fact that the regular args and the init args are separated by `--` should be transparent to the user.

Below is a POC of how it would look in Firecracker:

```rust
impl BootConfig {
    fn parse_boot_args(boot_args: &str) -> (&str, &str) {
        let params = boot_args.split("--").collect::<Vec<&str>>();

        let regular_params = params[0];
        let init_params = *(params.get(1)).unwrap_or(&"");

        return (regular_params, init_params)
    }

    /// Creates the BootConfig based on a given configuration.
    pub fn new(cfg: BootSourceConfig) -> std::result::Result<Self, BootSourceConfigError> {
        // ....
        let boot_args = match cfg.boot_args.as_ref() {
            None => DEFAULT_KERNEL_CMDLINE,
            Some(str) => str.as_str(),
        };
        let (regular_args, init_args) = parse_boot_args(boot_args);
        let mut cmdline = Cmdline::new(arch::CMDLINE_MAX_SIZE);
        cmdline
            .insert_str(regular_args)
            .map_err(|e| InvalidKernelCommandLine(e.to_string()))?;

        if !init_args.is_empty() {
            cmdline.insert_init_str(init_args);
        }

        // ....
    }
}
```

Then, in rust-vmm, we would need to add a function for `insert_init_str`. We can later maybe extend this to also support the equivalent of `insert`, but this is not needed at the moment. The `insert_init_str` should not contain the separator (` -- `). This should be added by the `Cmdline` when joining the `regular_args` with the `init_args`.

dianpopa:
Thanks @andreeaflorescu for the detailed example.

> I misunderstood the problem initially, but I looked at the bug report ([firecracker-microvm/firecracker#2709](https://github.com/firecracker-microvm/firecracker/issues/2709)) and things are a bit more clear now.
> 
> From the kernel command line point of view, `boot_args` does not have any meaning, so I do not think we should add it here. I initially thought that you want to add a function for specifying the `init` arguments.
> 
> I would propose we fix this problem using the following approach:
> 
> * In Firecracker, split the `boot_args` by `--`. This will give us the regular kernel parameters, and the init parameters.
> * In the `cmdline` utility, create 2 regions: one for regular parameters, and one for init parameters. This would follow the kernel command line specification, so it should be pretty straightforward to use. The `insert_init_str` should behave the same way as `insert_str`, just that it inserts the string in the `init` region instead of at the beginning. The fact that the regular args and the init args are separated by `--` should be transparent to the user.
> 
> Below is a POC of how it would look in Firecracker:
> 
> ```rust
> impl BootConfig {
>     fn parse_boot_args(boot_args: &str) -> (&str, &str) {
>         let params = boot_args.split("--").collect::<Vec<&str>>();
> 
>         let regular_params = params[0];
>         let init_params = *(params.get(1)).unwrap_or(&"");
> 
>         return (regular_params, init_params)
>     }
> 
>     /// Creates the BootConfig based on a given configuration.
>     pub fn new(cfg: BootSourceConfig) -> std::result::Result<Self, BootSourceConfigError> {
>         // ....
>         let boot_args = match cfg.boot_args.as_ref() {
>             None => DEFAULT_KERNEL_CMDLINE,
>             Some(str) => str.as_str(),
>         };
>         let (regular_args, init_args) = parse_boot_args(boot_args);
>         let mut cmdline = Cmdline::new(arch::CMDLINE_MAX_SIZE);
>         cmdline
>             .insert_str(regular_args)
>             .map_err(|e| InvalidKernelCommandLine(e.to_string()))?;
> 
>         if !init_args.is_empty() {
>             cmdline.insert_init_str(init_args);
>         }
> 
>         // ....
>     }
> }
> ```
> 
> Then, in rust-vmm, we would need to add a function for `insert_init_str`. We can later maybe extend this to also support the equivalent of `insert`, but this is not needed at the moment. The `insert_init_str` should not contain the separator (`--`). This should be added by the `Cmdline` when joining the `regular_args` with the `init_args`.


Your suggestions look good to me, however we might not be able to implement them exactly as proposed.
I understand you suggest to have 2 strings in the `Cmdline` implementation (one for init and one for regular params) and joining them in `as_str()`. However, this is not possible in Rust since it would mean to return a reference to a variable that is dynamically created in the `as_str` function:
```
 pub fn as_str(&self) -> &str {
     let cmdline = format!("{} -- {}", self.regular_params + self.init_params);
       cmdline.as_str()
```
The above does not compile in Rust. Let me know if I misunderstood the suggestion.

Since we do want to keep `as_str()` so that we do not break existing users, the only way is to alter `self.line` in `Cmdline`'s mutable functions by keeping track of the 2 regions with an index. This is not as straightforward as having 2 separate strings but I tried to keep the implementation as clean as possible. Take a look and let me know.

Not sure I understand the reference to how we should handle "--" in insert_init_args. We do need to take it into account when inserting a new param into the `init_params` regions since we do want to know if the `CMDLINE_SIZE` provided upon initialisation has space for that too, right?

PR #99, opened 2021-09-30
Implement `Debug` and `PartialEq` for `CmdLine`

### LINKED ISSUE #98: cmdline derive Debug & PartialEq
This is needed for tests.

alexandruag:
Fixes #98

### DIFF
--- a/src/cmdline/mod.rs
+++ b/src/cmdline/mod.rs
@@ -96,6 +96,7 @@ fn valid_element(s: &str) -> Result<()> {
 /// let cl_cstring = CString::new(cl).unwrap();
 /// assert_eq!(cl_cstring.to_str().unwrap(), "");
 /// ```
+#[derive(Debug)]
 pub struct Cmdline {
     line: String,
     capacity: usize,
@@ -337,6 +338,12 @@ impl From<Cmdline> for Vec<u8> {
     }
 }
 
+impl PartialEq for Cmdline {
+    fn eq(&self, other: &Self) -> bool {
+        self.line == other.line
+    }
+}
+
 #[cfg(test)]
 mod tests {
     use super::*;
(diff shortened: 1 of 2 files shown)


src/cmdline/mod.rs
```diff
@@ -337,6 +337,18 @@ impl From<Cmdline> for Vec<u8> {
     }
 }
 
+impl fmt::Debug for Cmdline {
```
andreeaflorescu, line 340: Why not use derive?

src/cmdline/mod.rs
```diff
@@ -337,6 +337,18 @@ impl From<Cmdline> for Vec<u8> {
...
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        write!(f, "{}", self.line)
+    }
+}
+
+impl PartialEq for Cmdline {
+    fn eq(&self, other: &Self) -> bool {
+        self.line == other.line
```
andreeaflorescu, line 348: Why not check the capacity as well? In which case we could also just derive partialeq.

src/cmdline/mod.rs
alexandruag: I liked this format better, but I've added a derive now.

src/cmdline/mod.rs
alexandruag: It looked more natural to not involve the capacity in checks for equality (i.e. the same as happens for `Vec` IIRC). I can also add another derive if there's a strong preference for that.

src/cmdline/mod.rs
```diff
@@ -479,4 +486,18 @@ mod tests {
         assert!(cl.insert_multiple("foo", &["bar", "baz"]).is_ok());
         assert_eq!(cl.as_str(), "foo=bar,baz");
     }
+
+    #[test]
+    fn test_partial_eq() {
+        let mut c1 = Cmdline::new(20);
+        let mut c2 = Cmdline::new(20);
```
lauralt, line 493: Can we add a test for different max sizes, but the same string?

alexandruag: Oh they were supposed to be different (i.e. 20 and 30), I've fixed this now.

lauralt (approved):

andreeaflorescu (approved):

PR #109, opened 2022-04-20
Make --no-default-features work 

uran0sH:
### Summary of the PR

If you want to use cmdline loader only, you can run build it with
--no-default-features

Signed-off-by: uran0sH <huangwenyuu@outlook.com>

### Requirements

requirements:

  description line.

### DIFF
--- a/.buildkite/custom-tests.json
+++ b/.buildkite/custom-tests.json
@@ -13,6 +13,14 @@
             "platform": [
                 "x86_64"
             ]
+        },
+        {
+            "test_name": "build-no-default-features",
+            "command": "cargo build --release --no-default-features",
+            "platform": [
+                "x86_64",
+                "aarch64"
+            ]
         }
     ]
 }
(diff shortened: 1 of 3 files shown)


Cargo.toml
```diff
@@ -12,10 +12,11 @@ readme = "README.md"
 autobenches = false
 
 [features]
-default = ["elf", "pe"]
+default = ["elf", "pe", "cmdline"]
 bzimage = []
 elf = []
 pe = []
+cmdline = []
```
andreeaflorescu, line 19: We should make it possible to build without any features. Adding `cmdline` as a feature is not necessary, and increases the complexity.

andreeaflorescu (changes requested):
We should also add a test to check that the PR actually fixes the bug. To do that, we should create a similar test to this one: https://github.com/rust-vmm/linux-loader/blob/3d80c65a62ef032a835332ddd64dddd913f641ec/.buildkite/custom-tests.json#L11. In the test we should run ```cargo build --no-default-features```.

andreeaflorescu (approved):

lauralt (approved):

PR #115, opened 2022-09-07
Cmdline refactoring

andreitraistaru:
### Summary of the PR

Reworking the Cmdline to leverage handling of boot args and init args. This PR should solve the #116 issue.

### Requirements

requirements:

  description line.
  test.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,20 @@
+# [Unreleased]
+
+## Added
+- Added `NoBootArgsInserted` error type used when calling `as_cstring` on a `Cmdline`
+  containing only init args.
+- Added `InvalidCapacity` error type used when trying to create a `Cmdline`
+  either with zero capacity (using `new` or `try_from`) or with smaller capacity
+  than required for arguments (using `try_from`).
+- Added `insert_init_args` method allowing insertion of init arguments into `Cmdline`.
+
+## Changed
+- Removed `InvalidDevice` error type (it wasn't used anywhere).
+- Replaced `From` with `TryFrom<Cmdline>` for `Vec<u8>` to be able
+  to propagate errors returned by `as_cstring` when converting a `Cmdline` to `Vec<u8>`.
+- Support added for both boot and init arguments in `try_from`.
+- Changed `new` to return `Result` for invalid command line capacity handling.
+
 # [v0.6.0]
 
 ## Changed
--- a/src/cmdline/mod.rs
+++ b/src/cmdline/mod.rs
@@ -14,15 +14,19 @@ use std::result;
 
 use vm_memory::{Address, GuestAddress, GuestUsize};
 
+const INIT_ARGS_SEPARATOR: &str = " -- ";
+
 /// The error type for command line building operations.
 #[derive(Debug, PartialEq, Eq)]
 pub enum Error {
     /// Null terminator identified in the command line.
     NullTerminator,
+    /// No boot args inserted into cmdline.
+    NoBootArgsInserted,
+    /// Invalid capacity provided.
+    InvalidCapacity,
     /// Operation would have resulted in a non-printable ASCII character.
     InvalidAscii,
-    /// Invalid device passed to the kernel command line builder.
-    InvalidDevice(String),
     /// Key/Value Operation would have had a space in it.
     HasSpace,
     /// Key/Value Operation would have had an equals sign in it.
@@ -41,12 +45,9 @@ impl fmt::Display for Error {
             Error::NullTerminator => {
                 write!(f, "Null terminator detected in the command line structure.")
             }
+            Error::NoBootArgsInserted => write!(f, "Cmdline cannot contain only init args."),
+            Error::InvalidCapacity => write!(f, "Invalid cmdline capacity provided."),
             Error::InvalidAscii => write!(f, "String contains a non-printable ASCII character."),
-            Error::InvalidDevice(ref dev) => write!(
-                f,
-                "Invalid device passed to the kernel command line builder: {}.",
-                dev
-            ),
             Error::HasSpace => write!(f, "String contains a space."),
             Error::HasEquals => write!(f, "String contains an equals sign."),
             Error::MissingVal(ref k) => write!(f, "Missing value for key {}.", k),
@@ -91,20 +92,20 @@ fn valid_element(s: &str) -> Result<()> {
 }
 
 /// A builder for a kernel command line string that validates the string as it's being built.
-/// A `CString` can be constructed from this directly using `CString::new`.
 ///
 /// # Examples
 ///
 /// ```rust
 /// # use linux_loader::cmdline::*;
 /// # use std::ffi::CString;
-/// let cl = Cmdline::new(100);
-/// let cl_cstring = CString::new(cl).unwrap();
-/// assert_eq!(cl_cstring.to_str().unwrap(), "");
+/// let mut cl = Cmdline::new(100).unwrap();
+/// cl.insert_str("foobar").unwrap();
+/// assert_eq!(cl.as_cstring().unwrap().as_bytes_with_nul(), b"foobar\0");
 /// ```
 #[derive(Clone, Debug)]
 pub struct Cmdline {
-    line: String,
+    boot_args: String,
+    init_args: String,
     capacity: usize,
 }
 
@@ -119,39 +120,23 @@ impl Cmdline {
     ///
     /// ```rust
     /// # use linux_loader::cmdline::*;
-    /// let cl = Cmdline::new(100);
+    /// let cl = Cmdline::new(100).unwrap();
     /// ```
     /// [`Cmdline`]: struct.Cmdline.html
-    pub fn new(capacity: usize) -> Cmdline {
-        assert_ne!(capacity, 0);
-        Cmdline {
-            line: String::with_capacity(capacity),
+    pub fn new(capacity: usize) -> Result<Cmdline> {
+        if capacity == 0 {
+            return Err(Error::InvalidCapacity);
+        }
+
+        Ok(Cmdline {
+            boot_args: String::new(),
+            init_args: String::new(),
             capacity,
-        }
+        })
     }
 
-    fn has_capacity(&self, more: usize) -> Result<()> {
-        let needs_space = if self.line.is_empty() { 0 } else { 1 };
-        if self.line.len() + more + needs_space < self.capacity {
-            Ok(())
-        } else {
-            Err(Error::TooLarge)
-        }
-    }
-
-    fn start_push(&mut self) {
-        if !self.line.is_empty() {
-            self.line.push(' ');
-        }
-    }
-
-    fn end_push(&mut self) {
-        // This assert is always true because of the `has_capacity` check that each insert method
-        // uses.
-        assert!(self.line.len() < self.capacity);
-    }
-
-    /// Validates and inserts a key-value pair into this command line.
+    /// Validates and inserts a key-value pair representing a boot
+    /// arg of the command line.
     ///
     /// # Arguments
     ///
@@ -162,7 +147,7 @@ impl Cmdline {
     ///
     /// ```rust
     /// # use linux_loader::cmdline::*;
-    /// let mut cl = Cmdline::new(100);
+    /// let mut cl = Cmdline::new(100).unwrap();
     /// cl.insert("foo", "bar");
     /// assert_eq!(cl.as_cstring().unwrap().as_bytes_with_nul(), b"foo=bar\0");
     /// ```
@@ -172,18 +157,14 @@ impl Cmdline {
 
         valid_element(k)?;
         valid_element(v)?;
-        self.has_capacity(k.len() + v.len() + 1)?;
 
-        self.start_push();
-        self.line.push_str(k);
-        self.line.push('=');
-        self.line.push_str(v);
-        self.end_push();
+        let kv_str = format!("{}={}", k, v);
 
-        Ok(())
+        self.insert_str(kv_str)
     }
 
-    /// Validates and inserts a key-value1,...,valueN pair into this command line.
+    /// Validates and inserts a key-value1,...,valueN pair representing a
+    /// boot arg of the command line.
     ///
     /// # Arguments
     ///
@@ -195,10 +176,12 @@ impl Cmdline {
     /// ```rust
     /// # use linux_loader::cmdline::*;
     /// # use std::ffi::CString;
-    /// let mut cl = Cmdline::new(100);
+    /// let mut cl = Cmdline::new(100).unwrap();
     /// cl.insert_multiple("foo", &["bar", "baz"]);
-    /// let cl_cstring = CString::new(cl).unwrap();
-    /// assert_eq!(cl_cstring.to_str().unwrap(), "foo=bar,baz");
+    /// assert_eq!(
+    ///     cl.as_cstring().unwrap().as_bytes_with_nul(),
+    ///     b"foo=bar,baz\0"
+    /// );
     /// ```
     pub fn insert_multiple<T: AsRef<str>>(&mut self, key: T, vals: &[T]) -> Result<()> {
         let k = key.as_ref();
@@ -223,7 +206,8 @@ impl Cmdline {
         self.insert_str(kv_str)
     }
 
-    /// Validates and inserts a string to the end of the current command line.
+    /// Inserts a string in the boot args; returns an error if the string
+    /// is invalid.
     ///
     /// # Arguments
     ///
@@ -234,22 +218,108 @@ impl Cmdline {
     /// ```rust
     /// # use linux_loader::cmdline::*;
     /// # use std::ffi::CString;
-    /// let mut cl = Cmdline::new(100);
-    /// cl.insert_str("foobar");
-    /// let cl_cstring = CString::new(cl).unwrap();
-    /// assert_eq!(cl_cstring.to_str().unwrap(), "foobar");
+    /// let mut cl = Cmdline::new(100).unwrap();
+    /// cl.insert_str("foobar").unwrap();
+    /// assert_eq!(cl.as_cstring().unwrap().as_bytes_with_nul(), b"foobar\0");
     /// ```
     pub fn insert_str<T: AsRef<str>>(&mut self, slug: T) -> Result<()> {
-        let s = slug.as_ref();
+        // Step 1: Check if the string provided is a valid boot arg string and remove any
+        // leading or trailing whitespaces.
+        let s = slug.as_ref().trim();
         valid_str(s)?;
 
-        self.has_capacity(s.len())?;
+        // Step 2: Check if cmdline capacity is not exceeded when inserting the boot arg
+        // string provided.
+        let mut cmdline_size = self.get_null_terminated_representation_size();
 
-        self.start_push();
-        self.line.push_str(s);
-        self.end_push();
+        // Count extra space required if this is not the first boot arg of the cmdline.
+        if !self.boot_args.is_empty() {
+            cmdline_size = cmdline_size.checked_add(1).ok_or(Error::TooLarge)?;
+        }
+
+        // Count extra space required for the insertion of the new boot arg string.
+        cmdline_size = cmdline_size.checked_add(s.len()).ok_or(Error::TooLarge)?;
+
+        if cmdline_size > self.capacity {
+            return Err(Error::TooLarge);
+        }
+
+        // Step 3: Insert the string as boot args to the cmdline.
+        if !self.boot_args.is_empty() {
+            self.boot_args.push(' ');
+        }
+
+        self.boot_args.push_str(s);
 
         Ok(())
+    }
+
+    /// Inserts a string in the init args; returns an error if the string
+    /// is invalid.
+    ///
+    /// # Arguments
+    ///
+    /// * `slug` - String to be appended to the command line.
+    ///
+    /// # Examples
+    ///
+    /// ```rust
+    /// # use linux_loader::cmdline::*;
+    /// # use std::ffi::CString;
+    /// let mut cl = Cmdline::new(100).unwrap();
+    /// cl.insert_str("foo").unwrap();
+    /// cl.insert_init_args("bar").unwrap();
+    /// assert_eq!(
+    ///     cl.as_cstring().unwrap().as_bytes_with_nul(),
+    ///     b"foo -- bar\0"
+    /// );
+    /// ```
+    pub fn insert_init_args<T: AsRef<str>>(&mut self, slug: T) -> Result<()> {
+        // Step 1: Check if the string provided is a valid init arg string and remove any
+        // leading or trailing whitespaces.
+        let s = slug.as_ref().trim();
+        valid_str(s)?;
+
+        // Step 2: Check if cmdline capacity is not exceeded when inserting the init arg
+        // string provided.
+        let mut cmdline_size = self.get_null_terminated_representation_size();
+
+        // Count extra space required if this is not the first init arg of the cmdline.
+        cmdline_size = cmdline_size
+            .checked_add(if self.init_args.is_empty() {
+                INIT_ARGS_SEPARATOR.len()
+            } else {
+                1
+            })
+            .ok_or(Error::TooLarge)?;
+
+        // Count extra space required for the insertion of the new init arg string.
+        cmdline_size = cmdline_size.checked_add(s.len()).ok_or(Error::TooLarge)?;
+
+        if cmdline_size > self.capacity {
+            return Err(Error::TooLarge);
+        }
+
+        // Step 3: Insert the string as init args to the cmdline.
+        if !self.init_args.is_empty() {
+            self.init_args.push(' ');
+        }
+
+        self.init_args.push_str(s);
+
+        Ok(())
+    }
+
+    fn get_null_terminated_representation_size(&self) -> usize {
+        // Counting current size of the cmdline (no overflows are possible as long as the cmdline
+        // size is always smaller or equal to the cmdline capacity provided in constructor)
+        let mut cmdline_size = self.boot_args.len() + 1; // for null terminator
+
+        if !self.init_args.is_empty() {
+            cmdline_size += INIT_ARGS_SEPARATOR.len() + self.init_args.len();
+        }
+
+        cmdline_size
     }
 
     /// Returns a C compatible representation of the command line
@@ -263,12 +333,28 @@ impl Cmdline {
     ///
     /// ```rust
     /// # use linux_loader::cmdline::*;
-    /// let mut cl = Cmdline::new(10);
-    /// cl.insert_str("foobar");
-    /// assert_eq!(cl.as_cstring().unwrap().as_bytes_with_nul(), b"foobar\0");
+    /// let mut cl = Cmdline::new(20).unwrap();
+    /// cl.insert_str("foo").unwrap();
+    /// cl.insert_init_args("bar").unwrap();
+    /// assert_eq!(
+    ///     cl.as_cstring().unwrap().as_bytes_with_nul(),
+    ///     b"foo -- bar\0"
+    /// );
     /// ```
     pub fn as_cstring(&self) -> Result<CString> {
-        CString::new(self.line.to_string()).map_err(|_| Error::NullTerminator)
+        if self.boot_args.is_empty() && self.init_args.is_empty() {
+            CString::new("".to_string()).map_err(|_| Error::NullTerminator)
+        } else if self.boot_args.is_empty() {
+            Err(Error::NoBootArgsInserted)
+        } else if self.init_args.is_empty() {
+            CString::new(self.boot_args.to_string()).map_err(|_| Error::NullTerminator)
+        } else {
+            CString::new(format!(
+                "{}{}{}",
+                self.boot_args, INIT_ARGS_SEPARATOR, self.init_args
+            ))
+            .map_err(|_| Error::NullTerminator)
+        }
     }
 
     /// Adds a virtio MMIO device to the kernel command line.
@@ -281,10 +367,10 @@ impl Cmdline {
     ///
     /// # Arguments
     ///
-    /// * `size` - size of the slot the device occupies on the MMIO bus.
-    /// * `baseaddr` - physical base address of the device.
-    /// * `irq` - interrupt number to be used by the device.
-    /// * `id` - optional platform device ID.
+    /// * `size` - Size of the slot the device occupies on the MMIO bus.
+    /// * `baseaddr` - Physical base address of the device.
+    /// * `irq` - Interrupt number to be used by the device.
+    /// * `id` - Optional platform device ID.
     ///
     /// # Examples
     ///
@@ -292,13 +378,12 @@ impl Cmdline {
     /// # use linux_loader::cmdline::*;
     /// # use std::ffi::CString;
     /// # use vm_memory::{GuestAddress, GuestUsize};
-    /// let mut cl = Cmdline::new(100);
+    /// let mut cl = Cmdline::new(100).unwrap();
     /// cl.add_virtio_mmio_device(1 << 12, GuestAddress(0x1000), 5, Some(42))
     ///     .unwrap();
-    /// let cl_cstring = CString::new(cl).unwrap();
     /// assert_eq!(
-    ///     cl_cstring.to_str().unwrap(),
-    ///     "virtio_mmio.device=4K@0x1000:5:42"
+    ///     cl.as_cstring().unwrap().as_bytes_with_nul(),
+    ///     b"virtio_mmio.device=4K@0x1000:5:42\0"
     /// );
     /// ```
     ///
@@ -343,17 +428,107 @@ impl Cmdline {
         }
         size.to_string()
     }
+
+    fn check_outside_double_quotes(slug: &str) -> bool {
+        slug.matches('\"').count() % 2 == 0
+    }
+
+    /// Tries to build a [`Cmdline`] with a given capacity from a str. The format of the
+    /// str provided must be one of the followings:
+    /// -> <boot args> -- <init args>
+    /// -> <boot args>
+    /// where <boot args> and <init args> can contain '--' only if double quoted and
+    /// <boot args> and <init args> contain at least one non-whitespace char each.
+    ///
+    /// Providing a str not following these rules might end up in undefined behaviour of
+    /// the resulting `Cmdline`.
+    ///
+    /// # Arguments
+    ///
+    /// * `cmdline_raw` - Contains boot params and init params of the cmdline.
+    /// * `capacity` - Capacity of the cmdline.
+    ///
+    /// # Examples
+    ///
+    /// ```rust
+    /// # use linux_loader::cmdline::*;
+    /// let cl = Cmdline::try_from("foo -- bar", 100).unwrap();
+    /// assert_eq!(
+    ///     cl.as_cstring().unwrap().as_bytes_with_nul(),
+    ///     b"foo -- bar\0"
+    /// );
+    /// ```
+    pub fn try_from(cmdline_raw: &str, capacity: usize) -> Result<Cmdline> {
+        // The cmdline_raw argument should contain no more than one INIT_ARGS_SEPARATOR sequence
+        // that is not double quoted; in case the INIT_ARGS_SEPARATOR is found all chars following
+        // it will be parsed as init args.
+
+        if capacity == 0 {
+            return Err(Error::InvalidCapacity);
+        }
+
+        // Step 1: Extract boot args and init args from input by searching for INIT_ARGS_SEPARATOR.
+
+        // Check first occurrence of the INIT_ARGS_SEPARATOR that is not between double quotes.
+        // All chars following the INIT_ARGS_SEPARATOR will be parsed as init args.
+        let (mut boot_args, mut init_args) = match cmdline_raw
+            .match_indices(INIT_ARGS_SEPARATOR)
+            .find(|&separator_occurrence| {
+                Self::check_outside_double_quotes(&cmdline_raw[..(separator_occurrence.0)])
+            }) {
+            None => (cmdline_raw, ""),
+            Some((delimiter_index, _)) => (
+                &cmdline_raw[..delimiter_index],
+                // This does not overflow as long as `delimiter_index + INIT_ARGS_SEPARATOR.len()`
+                // is pointing to the first char after the INIT_ARGS_SEPARATOR which always exists;
+                // as a result, `delimiter_index + INIT_ARGS_SEPARATOR.len()` is less or equal to the
+                // length of the initial string.
+                &cmdline_raw[(delimiter_index + INIT_ARGS_SEPARATOR.len())..],
+            ),
+        };
+
+        boot_args = boot_args.trim();
+        init_args = init_args.trim();
+
+        // Step 2: Check if capacity provided for the cmdline is not exceeded and create a new `Cmdline`
+        // if size check passes.
+        let mut cmdline_size = boot_args.len().checked_add(1).ok_or(Error::TooLarge)?;
+
+        if !init_args.is_empty() {
+            cmdline_size = cmdline_size
+                .checked_add(INIT_ARGS_SEPARATOR.len())
+                .ok_or(Error::TooLarge)?;
+
+            cmdline_size = cmdline_size
+                .checked_add(init_args.len())
+                .ok_or(Error::TooLarge)?;
+        }
+
+        if cmdline_size > capacity {
+            return Err(Error::InvalidCapacity);
+        }
+
+        Ok(Cmdline {
+            boot_args: boot_args.to_string(),
+            init_args: init_args.to_string(),
+            capacity,
+        })
+    }
 }
 
-impl From<Cmdline> for Vec<u8> {
-    fn from(cmdline: Cmdline) -> Vec<u8> {
-        cmdline.line.into_bytes()
+impl TryFrom<Cmdline> for Vec<u8> {
+    type Error = Error;
+
+    fn try_from(cmdline: Cmdline) -> result::Result<Self, Self::Error> {
+        cmdline
+            .as_cstring()
+            .map(|cmdline_cstring| cmdline_cstring.into_bytes_with_nul())
     }
 }
 
 impl PartialEq for Cmdline {
     fn eq(&self, other: &Self) -> bool {
-        self.line == other.line
+        self.as_cstring() == other.as_cstring()
     }
 }
 
@@ -362,23 +537,22 @@ mod tests {
     use super::*;
     use std::ffi::CString;
 
+    const CMDLINE_MAX_SIZE: usize = 4096;
+
     #[test]
     fn test_insert_hello_world() {
-        let mut cl = Cmdline::new(100);
+        let mut cl = Cmdline::new(100).unwrap();
         assert_eq!(cl.as_cstring().unwrap().as_bytes_with_nul(), b"\0");
         assert!(cl.insert("hello", "world").is_ok());
         assert_eq!(
             cl.as_cstring().unwrap().as_bytes_with_nul(),
             b"hello=world\0"
         );
-
-        let s = CString::new(cl).expect("failed to create CString from Cmdline");
-        assert_eq!(s, CString::new("hello=world").unwrap());
     }
 
     #[test]
     fn test_insert_multi() {
-        let mut cl = Cmdline::new(100);
+        let mut cl = Cmdline::new(100).unwrap();
         assert!(cl.insert("hello", "world").is_ok());
         assert!(cl.insert("foo", "bar").is_ok());
         assert_eq!(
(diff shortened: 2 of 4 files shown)


src/cmdline/mod.rs
```diff
@@ -116,134 +97,98 @@ impl Cmdline {
...
-    pub fn new(capacity: usize) -> Cmdline {
-        assert_ne!(capacity, 0);
+    pub fn new(cmdline_capacity: usize) -> Cmdline {
+        assert!(cmdline_capacity > CMDLINE_NO_OF_AUXILIARY_CHARS);
+
         Cmdline {
-            line: String::with_capacity(capacity),
-            capacity,
+            boot_args: String::with_capacity(cmdline_capacity),
+            init_args: String::with_capacity(cmdline_capacity),
```
aghecenco, line 105: This allows both components of the command line to have the maximum capacity, which would violate the capacity constraint on the entire `Cmdline` object. I don't think we can preallocate space for `String`s here anymore.

src/cmdline/mod.rs
```diff
@@ -72,18 +61,6 @@ fn valid_str(s: &str) -> Result<()> {
     }
 }
 
-fn valid_element(s: &str) -> Result<()> {
```
aghecenco, line 75: Why has this been removed? Components of the boot args (keys, corresponding values and standalone boot args) should still be validated.

src/cmdline/mod.rs
```diff
@@ -116,134 +97,98 @@ impl Cmdline {
...
-    /// # use std::ffi::CString;
-    /// let mut cl = Cmdline::new(100);
-    /// cl.insert_str("foobar");
+    /// cl.insert_init_arg("foobar").expect("");
     /// let cl_cstring = CString::new(cl).unwrap();
     /// assert_eq!(cl_cstring.to_str().unwrap(), "foobar");
     /// ```
-    pub fn insert_str<T: AsRef<str>>(&mut self, slug: T) -> Result<()> {
-        let s = slug.as_ref();
+    pub fn insert_init_arg<T: AsRef<str>>(&mut self, slug: T) -> Result<()> {
```
aghecenco, line 174: This function looks a lot like the current body of `insert_boot_arg`, which in turn looks somewhat like the old `insert_str`, so I would keep 1 single function for inserting a plain string alongside key-value insertion functions for boot args only.

src/cmdline/mod.rs
```diff
@@ -336,101 +287,88 @@ impl Cmdline {
     }
 }
 
+impl TryFrom<(usize, String)> for Cmdline {
+    type Error = ();
+
+    fn try_from(
+        (cmdline_capacity, cmdline_raw): (usize, String),
+    ) -> result::Result<Self, Self::Error> {
+        let (boot_args, init_args) = cmdline_raw.split_once("-- ").unwrap();
```
aghecenco: Don't `unwrap` in a `TryFrom`, propagate the error instead.

src/cmdline/mod.rs
```diff
@@ -336,101 +287,88 @@ impl Cmdline {
...
+
+        if boot_args.len() + init_args.len() + CMDLINE_NO_OF_AUXILIARY_CHARS > cmdline_capacity {
+            return Err(());
```
aghecenco: This should be a specific error, not `()`.

src/cmdline/mod.rs
```diff
@@ -336,101 +287,88 @@ impl Cmdline {
...
-    fn test_insert_hello_world() {
-        let mut cl = Cmdline::new(100);
-        assert_eq!(cl.as_str(), "");
-        assert!(cl.insert("hello", "world").is_ok());
-        assert_eq!(cl.as_str(), "hello=world");
-
-        let s = CString::new(cl).expect("failed to create CString from Cmdline");
-        assert_eq!(s, CString::new("hello=world").unwrap());
-    }
+    const DEFAULT_CMDLINE_RAW : &str = "console=ttyS0 nomodules random.trust_cpu=on reboot=k panic=1 pci=off ip=:::::eth0:dhcp init=/bin/cat -- /etc/password";
```
aghecenco, line 334: This is the same as `DEFAULT_CMDLINE_BOOT_ARGS` plus an init arg so I think it should be defined as such.

aghecenco, line 334: I don't think the *default* command line we test with needs to be this explicit. Several of these boot args bring no additional value - from a testing standpoint, all key-value args are the same.

aghecenco (dismissed):
There are some breaking changes I disagree with, and I left comments throughout the code to explain. I also have some style nits but those can wait until later.
I also have some commit-specific comments:
- Commit 1caf1a7f4b15ea57e9f751786544f6ffe2ad63af :  typo `abe/able`
- Commit 9dbaaf5c6e2a63991730022bc6a19247332b2929 should not exist. It looks like its purpose is to test integration with a consumer of this crate. Such changes are dev-only and should be made in the vmm that consumes this crate, not the other way around.
- Commit c52ba27f801bed35faff9d51f58bc30b63588f61 should not exist either, tests should pass on each commit.

aghecenco (commented):
@andreitraistaru  additionally, as this PR fixes a limitation (essentially a bug) - namely, that `Cmdline` can't handle init args -, I think you should open an issue for that and assign the PR to it.

src/cmdline/mod.rs
```diff
@@ -116,134 +97,98 @@ impl Cmdline {
     /// let cl = Cmdline::new(100);
     /// ```
     /// [`Cmdline`]: struct.Cmdline.html
-    pub fn new(capacity: usize) -> Cmdline {
-        assert_ne!(capacity, 0);
+    pub fn new(cmdline_capacity: usize) -> Cmdline {
+        assert!(cmdline_capacity > CMDLINE_NO_OF_AUXILIARY_CHARS);
```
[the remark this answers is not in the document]
andreeaflorescu: Generally speaking we should never use asserts in library code for invalid input. Asserts in library should be reserved for programming errors. In this case, this is a user error for which we should return an error.

src/cmdline/mod.rs
andreeaflorescu: If we remove the cmdline_capacity, the code is going to be greatly simplified. We could have a `MAX_SIZE` to make sure that the cmdline is still within reason to protect against potential malicious actors.

src/cmdline/mod.rs
```diff
@@ -254,12 +199,18 @@ impl Cmdline {
...
+    /// assert!(cl.insert_boot_arg("foobar").is_ok());
     /// assert_eq!(cl.as_str(), "foobar");
     /// ```
-    pub fn as_str(&self) -> &str {
-        self.line.as_str()
+    pub fn as_str(&self) -> String {
+        if self.init_args.is_empty() {
+            self.boot_args.to_string()
+        } else if self.boot_args.is_empty() {
+            format!("-- {}", self.init_args)
```
andreeaflorescu, line 210: Would this be a valid kernel command line? I wonder if we should return an error in this case.

andreeaflorescu (commented):
There are some high level changes required to the PR, I did not review the tests and in depth the TryFrom functionality as they're likely to change once a newer revision is pushed.

andreitraistaru:
Thank you so much for your implication and help! I'll take each comment and address it accordingly

src/cmdline/mod.rs
andreitraistaru: I though about preallocating cmdline_capacity for both boot_args and init_args to simplify the implementation. Implementing like this guarantees that we have enough space even though no init arg will be provided (and thus, the entire space will be used for boot_args). We will NOT exceed cmdline_capacity when joining boot_args and init_args because before each insert in the Cmdline, we check the condition for capacity of the joined cmdline (the function has_capacity checks for the sum of lengths of boot_args and init_args). Yes, we can definitely implement without preallocations and keep the same checks - the only disadvantage that I can see for this approach would be the overhead introduced by repetitive inserts into string, but probably this would be insignificant as long as there are few inserts for the Cmdline. I wouldn't remove the cmdline_capacity from the constructor parameters as long as this is a constant specific to the architecture and I consider it part of the abstractions the rust-vmm/linux-loader imposes.

src/cmdline/mod.rs
```diff
@@ -116,134 +97,98 @@ impl Cmdline {
...
     /// # use linux_loader::cmdline::*;
     /// # use std::ffi::CString;
     /// let mut cl = Cmdline::new(100);
-    /// cl.insert("foo", "bar");
+    /// cl.insert_boot_arg("foobar").expect("");
     /// let cl_cstring = CString::new(cl).unwrap();
-    /// assert_eq!(cl_cstring.to_str().unwrap(), "foo=bar");
+    /// assert_eq!(cl_cstring.to_str().unwrap(), "foobar");
     /// ```
-    pub fn insert<T: AsRef<str>>(&mut self, key: T, val: T) -> Result<()> {
```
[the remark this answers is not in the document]
andreitraistaru: Ok, that was one aspect I was not sure about. I saw that we don't use it anywhere in Firecracker so I decided to remove it. However, from the abstraction point of view, yes... you are right... it would be better to keep this feature available. I'll add the methods related to key/value back into the commit.

src/cmdline/mod.rs
```diff
@@ -116,134 +97,98 @@ impl Cmdline {
...
     ///
     /// ```rust
     /// # use linux_loader::cmdline::*;
     /// # use std::ffi::CString;
     /// let mut cl = Cmdline::new(100);
-    /// cl.insert_multiple("foo", &["bar", "baz"]);
-    /// let cl_cstring = CString::new(cl).unwrap();
-    /// assert_eq!(cl_cstring.to_str().unwrap(), "foo=bar,baz");
-    /// ```
-    pub fn insert_multiple<T: AsRef<str>>(&mut self, key: T, vals: &[T]) -> Result<()> {
```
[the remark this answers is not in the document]
andreitraistaru: yes, I'll rollback this method as well.

src/cmdline/mod.rs
andreitraistaru: it was removed because this method was used only inside the methods removed (and mentioned above). Considering the removal of these methods the valid_element was no longer needed. Considering the comments above, I'll rollback this method as well.

src/cmdline/mod.rs
andreitraistaru: Ok, so if I got it right you suggest to keep 2 functions: one for inserting a string into init_args and the other one for inserting a key-value pair into boot_args. To make it clear for the user, I would suggest to have a naming like `insert_init_args_str()` and `insert_boot_args_pair()`. Do you agree with this naming convention? Otherwise, it would be a little bit confusing for the user, even though the types of the parameters of the functions clarifies the ambiguity.

src/cmdline/mod.rs
andreitraistaru: Ok, it makes sense... as long as the method's name starts with 'try'.

src/cmdline/mod.rs
andreitraistaru: Regarding String::with_capacity(cmdline_capacity) I propose to declare it as String::new() and not provide any initial capacity at all. Before any insertion, we perform capacity checks considering both boot args and init args (if exists) such that insertion of too large args to be impossible. Do you have any concern with this approach?

src/cmdline/mod.rs
andreitraistaru: Also, I consider useful to keep a function that inserts a string into boot args as long as we might need to append boot args that are not key-value pairs.

src/cmdline/mod.rs
andreitraistaru: Regarding String::with_capacity(cmdline_capacity) I propose to declare it as String::new() and not provide any initial capacity at all. Before any insertion, we perform capacity checks considering both boot args and init args (if exists) such that insertion of too large args to be impossible. Do you have any concern with this approach?

andreitraistaru: Regarding the usecase in which we return format!("-- {}", self.init_args), I don't see it as an error because the method is called as_str so it does not imply any validity checks done by the cmdline; it just implies a representation of the current state of the cmdline; furthermore, we do not perform any validity checks of the boot arguments the user can insert into the cmdline either (apart from the basic check to make sure it's an ASCII string and it doesn't contain any space); thus, it's the responsibility of the user to provide correct data, we just offer the data structure for implementing a Cmdline.

src/cmdline/mod.rs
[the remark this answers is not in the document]
andreeaflorescu: This is not only used by Firecracker, but other customers as well. We shouldn't remove existing functionality.

src/cmdline/mod.rs
andreeaflorescu: Since we already have breaking changes at the interface level, I was thinking about making a more clear separation of concerns between bootargs and initargs by creating 2 separate structures to handle each. For the operations that are the same for both of them, we can re-use the code by implementing the same trait for both structures. For the functionality that is different, we need to define it in the corresponding structure. One example for separate functionality is the `insert_virtio` function which needs to be implemented only for `BootArgs`.

This will result in more code, but I think it makes the interface more logical as well.

```rust
pub struct BootArgs(String);
pub struct InitArgs(String);

pub trait UpdateArgs {
    fn insert_key_value<T: AsRef<str>>(&mut self, key: T, val: T) -> Result<()>;
    fn insert_multiple<T: AsRef<str>>(&mut self, key: T, vals: &[T]) -> Result<()>;
    ......
    /// we need to add all the other common functions here as well.
}

/// To reuse the code we can implement the trait for String and then just have blanket implementations for both InitArgs and BootArgs.
/// There are probably better ways to auto-implement the trait, but we need to think about this a bit more.

#[derive(Clone, Debug)]
pub struct Cmdline {
    boot_args: BootArgs,
    init_args: InitArgs,
}

impl UpdateArgs for BootArgs {...}
impl UpdateArgs for InitArgs {...}

impl BootArgs {
    pub fn insert_virtio_device.....
    /// here we need to add any other functions that are specific to the boot args.
.....
}

impl Cmdline {
     pub fn boot_args_mut(&mut self) -> &mut BootArgs {....}
     pub fn init_args_mut(&mut self) -> &mut InitArgs {...}
}

......
fn how_to_use() {
   let mut cmdline = Cmdline::try_from("some string");
   cmdline.boot_args_mut().insert_virtio_device(.....);
   cmdline.boot_args_mut().insert_multiple(.....);
}



```

src/cmdline/mod.rs
andreeaflorescu: The meaning of `as_str` is that you can pass this to the kernel, so in that sense the `Cmdline` will not get any more pushes. We are working with a fail fast methodology. That means that if there are error cases that we can check ourselves in a reasonable way, we should do that. For example, that is why are checking for ASCII only strings. We could also not check, and let the Linux Kernel fail to boot. But that would mean that we are passing erroneous data down to the lower levels for no good reason. Hence, my question whether Linux accepts a command line that only has init args.

andreitraistaru: Yes, you are right somehow it's always better to catch errors as early as possible. I'll investigate this usecase and come back with a result.

andreeaflorescu:
@andreitraistaru why are you removing the coverage files?

src/cmdline/mod.rs
andreitraistaru: Working on this

andreitraistaru:
@andreeaflorescu sorry, I mistakenly deleted those files.... rollbacking them now. thanks

andreitraistaru:
I did the separation of concerns with BootArgs and InitArgs as discussed previously; now, going back to write some documentation and fix the tests to have all green in buildkite; I'll let you know when ready for a new review.

There is one concern from my point of view: the way we currently handle cmdline capacity checks; I implemented a simple check when inserting in BootArgs and InitArgs structs such that user will not be able to abuse the memory and a final check in the as_str() method where the total capacity should be less or equal to the capacity given when instantiating a Cmdline object. Is it ok like this? Did I miss something? It would have been way too complicated to check the total length of the Cmdline when inserting into a BootArgs struct so I consider that the way I implemented is a good compromise

andreitraistaru:
I did some code changes to handle double quotes handling as well.

src/cmdline/mod.rs
```diff
@@ -336,172 +250,641 @@ impl Cmdline {
     }
 }
 
+impl UpdateArgs for BootArgs {
+    fn has_capacity(&self, more: usize) -> bool {
```
andreeaflorescu, line 254: Why is this implemented for each type of Arg? We should be able to use the same implementation twice. For example, we could implement the trait for String, and then here just call `self.0.has_capacity()`. Same for InitArgs. This also applies to the other functions that are duplicated.

src/cmdline/mod.rs
```diff
@@ -219,49 +159,20 @@ impl Cmdline {
...
-    /// # use linux_loader::cmdline::*;
-    /// let mut cl = Cmdline::new(10);
-    /// cl.insert_str("foobar");
-    /// assert_eq!(cl.as_str(), "foobar");
-    /// ```
-    pub fn as_str(&self) -> &str {
-        self.line.as_str()
-    }
+/// An abstraction for the boot arguments of a kernel command line.
+#[derive(Clone, Debug)]
```
andreeaflorescu, line 171: We should follow the API guidelines and derive the other useful traits as well. Here are the Rust guidelines: https://rust-lang.github.io/api-guidelines/interoperability.html

src/cmdline/mod.rs
```diff
@@ -336,172 +250,641 @@ impl Cmdline {
...
+    /// * `cmdline_capacity` - Command line capacity. Must be greater than 0.
+    ///
+    /// # Examples
+    ///
+    /// ```rust
+    /// # use linux_loader::cmdline::*;
+    /// let cl = Cmdline::new(100);
+    /// ```
+    /// [`Cmdline`]: struct.Cmdline.html
+    pub fn new(cmdline_capacity: usize) -> Result<Cmdline> {
```
andreeaflorescu, line 354: With the current design of the API I don't think that a `cmdline_capacity` makes a lot of sense anymore. If we want to, we can give users the possibility to override the `MAX_CAPACITY` by having 2 contructors:
- one that takes a capacity as a parameter
- and one that doesn't take a capacity as a parameter, and just uses the defined DEFAULT instead.

src/cmdline/mod.rs
```diff
@@ -336,172 +250,641 @@ impl Cmdline {
...
+    /// assert_eq!(cl.as_str().unwrap(), "foo -- bar");
+    /// ```
+    pub fn as_str(&self) -> Result<String> {
+        if self.boot_args.0.is_empty() {
+            return Err(Error::NoBootArgsInserted);
+        }
+
+        if self.init_args.0.is_empty() {
+            if self.boot_args.0.len() + 1 > self.cmdline_capacity {
+                // need space for the null terminator
```
andreeaflorescu, line 413: In the documentation comment we're saying that we're not adding the nul terminator, which one is the correct comment?

andreeaflorescu (commented):
Part 1. Still need to look into try_from.

src/cmdline/mod.rs
```diff
@@ -336,172 +250,669 @@ impl Cmdline {
...
+
+fn check_outside_double_quotes(slug: &str) -> bool {
+    slug.matches('\"').count() % 2 == 0
+}
+
+impl TryFrom<(usize, String)> for Cmdline {
+    type Error = String;
+
+    fn try_from(
+        (cmdline_capacity, cmdline_raw): (usize, String),
```
andreeaflorescu, line 435: If we want to have a capacity as well then I think implementing `TryFrom` is no longer a good idea because we have to use a tuple, so it's no longer clear how to initialize the `Cmdline`. In case we want to keep the capacity, we should just have a function on `Cmdline` so that we can have as named parameters a `value` and a `capacity`. I would recommend thought to drop `capacity` altogether and just check (as you're already doing) that the capacity of the generated string is less than the maximum allowed one by the linux kernel.

src/cmdline/mod.rs
```diff
@@ -219,49 +159,20 @@ impl Cmdline {
...
-    /// ```
-    pub fn as_str(&self) -> &str {
-        self.line.as_str()
-    }
+/// An abstraction for the boot arguments of a kernel command line.
+#[derive(Clone, Debug)]
+pub struct BootArgs(String);
 
+impl BootArgs {
+    /// here we need to add any other functions that are specific to the boot args.
```
andreeaflorescu, line 175: This comment should be removed.

src/cmdline/mod.rs
```diff
@@ -283,10 +194,11 @@ impl Cmdline {
...
     /// # use std::ffi::CString;
     /// # use vm_memory::{GuestAddress, GuestUsize};
-    /// let mut cl = Cmdline::new(100);
-    /// cl.add_virtio_mmio_device(1 << 12, GuestAddress(0x1000), 5, Some(42))
+    /// let mut cl = Cmdline::new(100).unwrap();
+    /// cl.boot_args_mut()
+    ///     .add_virtio_mmio_device(1 << 12, GuestAddress(0x1000), 5, Some(42))
     ///     .unwrap();
-    /// let cl_cstring = CString::new(cl).unwrap();
+    /// let cl_cstring = CString::new(cl.as_str().unwrap()).unwrap();
```
andreeaflorescu, line 201: I don't think this needs to be a CString anymore.

src/cmdline/mod.rs
```diff
@@ -336,172 +250,669 @@ impl Cmdline {
...
+pub struct InitArgs(String);
+
+impl UpdateArgs for InitArgs {
+    fn has_capacity(&self, more: usize) -> bool {
+        let mut auxiliary_space = 0;
+
+        if !self.0.is_empty() {
+            // this is NOT the first arg so we need to add an extra space between existing
+            // args and the new args to be added
+            auxiliary_space += 1;
```
andreeaflorescu, line 295: Shouldn't it be enough to initialize this to 1 here instead of adding 1?

andreeaflorescu (commented):
Part 2, still need to look into a few corner cases.

src/cmdline/mod.rs
andreitraistaru: yes, you are right.... still not get used to rust .... I initially tried to move has_capacity()'s implementation into impl UpdateArgs block (similar to insert_key_value()) but it was actually impossible because the logic of this method needs reference to self object of the struct for which we need the trait to be implemented and then I just find as a solution to duplicate the code... Thank you for your advice! I moved the implementation into impl UpdateArgs for String and then just called the methods from there.

src/cmdline/mod.rs
andreitraistaru: Actually, my concern here is that if we keep the capacity parameter of the Cmdline constructor and change the default MAX_CAPACITY value that we use when calling has_capacity() method of the UpdateArgs trait, we still don't handle this aspect very well: because we allow appending to the boot args string until reaching MAX_CAPACITY and also allow appending adding init args until reaches another MAX_CAPACITY in the init args string. When, eventually, call as_string() we throw an error because the joined cmdline exceeds MAX_CAPACITY; the best way to handle this case would be that in has_capacity() method to check somehow the length of both boot args string and init args string but implementing something like this means keeping a reference to init args in the BootArgs struct and vice versa which would introduce unjustified complexity from my point of view. Let's remove cmdline_capacity param from cmdline constructor and just keep a default MAX_CAPACITY for checks before inserts (preventing abuse) and before as_str as long as exceeding MAX_CAPACITY would definitely result into a erroneous cmdline; I chose 4096 as MAX_CAPACITY as long as this is the max size of a valid kernel cmdline according to the [spec](https://www.kernel.org/doc/html/v4.14/admin-guide/kernel-parameters.html)

src/cmdline/mod.rs
andreitraistaru: Adding implementation for Eq, PartialEq, Ord, PartialOrd, Hash, Default and Dsiplay traits as well for both BootArgs and InitArgs. Same traits where implemented for Cmdline, most of them using #derive, and a few of them by providing actual implementation.

src/cmdline/mod.rs
andreitraistaru: We don't add any '\0' char per se when returning the String from as_str() method. However, we need to consider the space required for the nul terminator within the cmdline_capacity so that's why I wrote the if condition adding + 1 (and added a comment to be more clear) - actually it's the same condition used as before, but written in another way.

andreitraistaru: I'll come back to the old condition and remove the comment.

andreitraistaru: But your questions triggered another question in my mind: who should be responsible for adding the '\0' char to the cmdline ? I'm asking because from what I understand, currently, nobody does this... I checked the linux-loader/loader crate and the function load_cmdline() does not do this; it just takes the cmdline.as_str() turns it into bytes and then writes these bytes into guest's memory; is my understanding right? shouldn't we add a '\0' at the end in as_str() method to make sure the cmdline is nul terminated?

src/cmdline/mod.rs
```diff
@@ -336,172 +250,669 @@ impl Cmdline {
...
+        }
+
+        if self.init_args.0.is_empty() {
+            if self.boot_args.0.len() + 1 > self.cmdline_capacity {
+                // need space for the null terminator
+                Err(Error::TooLarge)
+            } else {
+                Ok(self.boot_args.0.to_string())
+            }
+        } else if self.boot_args.0.len() + self.init_args.0.len() + 5 > self.cmdline_capacity {
```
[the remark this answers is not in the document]
andreitraistaru: Done.

src/cmdline/mod.rs
andreitraistaru: Ok, reading the comments below, I'll change the code and remove cmdline_capacity from constructor.

src/cmdline/mod.rs
andreitraistaru: Ok, implementing this way. I think that the cmdline_capacity param came from the fact that each arch has different cmdline_max sizes; we can allow the maximum of these values (4096)

src/cmdline/mod.rs
andreitraistaru: ups...; done

src/cmdline/mod.rs
andreitraistaru: simplifying even further and removing mutability by refactoring into: let auxiliary_space = if self.is_empty() { 0 } else { 1 };

src/cmdline/mod.rs
andreeaflorescu: On a second thought, that where we get the null terminated string from. The reason why we're using CString here is that one, it was just very poorly documented.

src/cmdline/mod.rs
andreeaflorescu: We should be using CString to get a null terminated string.

src/cmdline/mod.rs
```diff
@@ -336,172 +269,736 @@ impl Cmdline {
...
+    /// # Examples
+    ///
+    /// ```rust
+    /// # use linux_loader::cmdline::*;
+    /// let mut cl = Cmdline::new().unwrap();
+    /// assert!(cl.boot_args_mut().insert_str("foo").is_ok());
+    /// assert!(cl.init_args_mut().insert_str("bar").is_ok());
+    /// assert_eq!(cl.as_str().unwrap(), "foo -- bar");
+    /// ```
+    pub fn as_str(&self) -> Result<String> {
```
andreeaflorescu, line 389: To have a null terminated string here we can return a `CString`.

src/cmdline/mod.rs
```diff
@@ -345,42 +290,216 @@ impl Cmdline {
...
+    }
+}
+
+impl TryFrom<String> for Cmdline {
+    type Error = String;
+
+    fn try_from(cmdline_raw: String) -> result::Result<Self, Self::Error> {
+        // Checking for ' -- ' sequence is required as long as a sequence like 'foo-- bar' should not be
+        // handled as a boot param 'foo' and an init arg 'bar'
+        let (mut boot_args, mut init_args) = match cmdline_raw.split_once(" -- ") {
```
andreeaflorescu: Can we have a define for " -- " instead? For example, we can call it INIT_ARG_DELIMITER. Then, we should be able to use INIT_ARG_DELIMITER.len() instead of magic numbers in the code below.

src/cmdline/mod.rs
```diff
@@ -345,42 +290,216 @@ impl Cmdline {
...
+                    }
+                }
+            }
+            Some((boot_args_section, init_args_section)) => (boot_args_section, init_args_section),
+        };
+
+        boot_args = boot_args.trim();
+        init_args = init_args.trim();
+
+        let mut more_space = 1; // required for null terminator of the cmdline
```
andreeaflorescu: It would make the code more readable if we instead define a `cmdline_size` instead of `more_space`. Then, the code will read more like a story:
- cmdline size initially is boot_args.len() + 1;
- if in the cmdline we also have `init_args`, then, the cmdline_size += init_args.len() + INIT_ARGS_DELIMITER.len().
- if the cmdline size is bigger than CMDLINE_MAXSIZE, then we return an error

src/cmdline/mod.rs
andreitraistaru: We already have such a define: INIT_ARGS_SEPARATOR; I don't see any magic number below apart some variations from the ' -- ' separator such as '-- ' or ' --' that are used to catch some corner cases (but their only usecase is in try_from() so I considered that is easier to see the actual value instead of another define which is used in only one place). Looks like this comment is older and I forgot to close it or something... ping me if I miss something

src/cmdline/mod.rs
andreitraistaru: updating that

src/cmdline/mod.rs
```diff
@@ -220,57 +168,51 @@ impl Cmdline {
...
-    /// assert_eq!(cl_cstring.to_str().unwrap(), "foobar");
-    /// ```
-    pub fn insert_str<T: AsRef<str>>(&mut self, slug: T) -> Result<()> {
-        let s = slug.as_ref();
+    /// * `slug` - String to be appended to the args string.
+    fn insert_string<T: AsRef<str>>(&mut self, slug: T) -> Result<()>;
+}
+
+impl UpdateArgs for String {
+    fn has_capacity(&self, more: usize) -> bool {
```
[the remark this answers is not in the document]
andreitraistaru: Yes, that was what I was concerned in one of the comments above. From my point of view the solution to keep track of the size of the init args string while adding a new boot arg (and vice versa) would introduce a great complexity in the code that is not justified (I suppose we need to add a reference to the init string inside the boot arg structure and vice versa to be able to have the check properly implemented at this level, but the design would be hard to maintain).

However, my solution was to keep has_capacity from `Args` in order to prevent abuse (not to allow adding more that MAX_CAPACITY bytes into any of the init args string or boot args string because that would definitely result in an error when calling as_cstring()) and the real check already exist in the as_cstring().

Another option would be to have a method under `Cmdline` that takes two params (one specifying the new arguments that needs to be appended and the second one to specify the type of param that we try to add). This method can check if there is still space (according to all corner cases) because `Cmdline` has access to both init and boot arg strings and after that we call the corresponding insert method to the `init args` or `boot args` structs. In other words, to implement a wrapper over insert* methods in `Cmdline`. Doing like this mens we no longer need methods `boot_args_mut()` and `init_args_mut()` from `Cmdline`.

Would you like to go with this last approach?

src/cmdline/mod.rs
```diff
@@ -90,117 +95,294 @@ fn valid_element(s: &str) -> Result<()> {
...
+    ///
+    /// ```rust
+    /// # use linux_loader::cmdline::*;
+    /// let mut cl = Cmdline::new(3).unwrap();
+    /// assert!(cl.insert_string("foo", CmdlineArgType::BootArg).is_err());
+    /// cl.set_capacity(4).unwrap();
+    /// assert!(cl.insert_string("foo", CmdlineArgType::BootArg).is_ok());
+    /// assert_eq!(cl.set_capacity(3), Err(Error::InvalidCapacity));
+    /// ```
+    pub fn set_capacity(&mut self, new_capacity: usize) -> Result<()> {
```
andreeaflorescu, line 277: What use case does this function serve? We can already set the capacity in the constructor. When would we need to update that?

src/cmdline/mod.rs
```diff
@@ -90,117 +95,294 @@ fn valid_element(s: &str) -> Result<()> {
...
+            }
+        };
+
+        cmdline_size += more;
+
+        cmdline_size <= self.capacity
     }
 
-    /// Validates and inserts a key-value pair into this command line.
+    /// Validates and inserts a key=value pair as a new cmdline argument.
```
andreeaflorescu, line 326: Why change key=value? The `key-value` is written like that on purpose as it represents a concept. In the documentation comments we should use concepts, not implementation details as a general rule.

src/cmdline/mod.rs
```diff
@@ -90,117 +95,294 @@ fn valid_element(s: &str) -> Result<()> {
...
     /// # use linux_loader::cmdline::*;
-    /// let mut cl = Cmdline::new(100);
-    /// cl.insert("foo", "bar");
+    /// let mut cl = Cmdline::new(10).unwrap();
+    /// cl.insert_key_value("foo", "bar", CmdlineArgType::BootArg)
+    ///     .unwrap();
     /// assert_eq!(cl.as_cstring().unwrap().as_bytes_with_nul(), b"foo=bar\0");
     /// ```
-    pub fn insert<T: AsRef<str>>(&mut self, key: T, val: T) -> Result<()> {
+    pub fn insert_key_value<T: AsRef<str>>(
```
andreeaflorescu, line 343: Why not leave it as insert? The fact that it's key-value is clear from the parameters that it receives.

src/cmdline/mod.rs
```diff
@@ -220,55 +402,41 @@ impl Cmdline {
...
-    /// let cl_cstring = CString::new(cl).unwrap();
-    /// assert_eq!(cl_cstring.to_str().unwrap(), "foobar");
+    /// let mut cl = Cmdline::new(10).unwrap();
+    /// cl.insert_string("foo=bar", CmdlineArgType::BootArg)
+    ///     .unwrap();
+    /// assert_eq!(cl.as_cstring().unwrap().as_bytes_with_nul(), b"foo=bar\0");
     /// ```
-    pub fn insert_str<T: AsRef<str>>(&mut self, slug: T) -> Result<()> {
-        let s = slug.as_ref();
+    pub fn insert_string<T: AsRef<str>>(
```
andreeaflorescu, line 424: Why change the name of this? It still receives something that can be a `str`, so the name was correct before.

andreeaflorescu (commented):
Let's chat offline more about how we can get this PR to a mergable state.

src/cmdline/mod.rs
andreitraistaru: removing this as discussed

src/cmdline/mod.rs
andreitraistaru: keeping as it was initially

src/cmdline/mod.rs
andreitraistaru: going back to the old naming

src/cmdline/mod.rs
andreitraistaru: going back to the old naming

src/cmdline/mod.rs
```diff
@@ -234,24 +218,95 @@ impl Cmdline {
...
     /// ```
     pub fn insert_str<T: AsRef<str>>(&mut self, slug: T) -> Result<()> {
-        let s = slug.as_ref();
+        let s = slug.as_ref().trim();
+        valid_str(s)?;
+
+        let mut cmdline_size = self.get_null_terminated_representation_size();
+
+        // Checking if one extra space needs to be added before insertion of a new boot arg
+        cmdline_size += if !self.boot_args.is_empty() { 1 } else { 0 };
```
lauralt, line 232: I see some unchecked operations in this PR, is it possible to have overflows? If not, would be nice to add comments about that.

src/cmdline/mod.rs
```diff
@@ -234,24 +218,95 @@ impl Cmdline {
...
+
+        // Adding extra size required for the insertion of `more` bytes as boot args
```
lauralt, line 234: Is `more` a reference to some parameter or?

andreitraistaru: it was just an outdated comment; thank you for notification!

src/cmdline/mod.rs
andreitraistaru: fixing that...

andreitraistaru:
I added new tests for the try_from() functionality and fixed some issue discovered meanwhile. This is one of the most error-prone sections of the code, as long as there are many corner cases hard to detect

src/cmdline/mod.rs
```diff
@@ -234,24 +218,117 @@ impl Cmdline {
...
+        cmdline_size =
+            match cmdline_size.checked_add(if !self.boot_args.is_empty() { 1 } else { 0 }) {
+                None => {
+                    return Err(Error::TooLarge);
+                }
+                Some(new_cmdline_size) => new_cmdline_size,
+            };
+
+        // Adding extra size required for the insertion of the new boot args
+        cmdline_size = match cmdline_size.checked_add(s.len()) {
```
andreeaflorescu, line 241: We should convert Options into Results for neat code:
```rust
cmdline_size = cmdline_size.checked_add(s.len()).ok_or(Error::TooLarge)?;
```

src/cmdline/mod.rs
```diff
@@ -234,24 +218,117 @@ impl Cmdline {
...
-        let s = slug.as_ref();
+        let s = slug.as_ref().trim();
         valid_str(s)?;
 
-        self.has_capacity(s.len())?;
+        let mut cmdline_size = self.get_null_terminated_representation_size();
+
+        // Checking if one extra space needs to be added before insertion of a new boot arg
+        cmdline_size =
+            match cmdline_size.checked_add(if !self.boot_args.is_empty() { 1 } else { 0 }) {
```
andreeaflorescu, line 233: We should have the if first becasue we shouldn't have to add 0 to size. We can also use `ok_or` to return an error in case of an overflow.

andreitraistaru: done

src/cmdline/mod.rs
andreitraistaru: done

src/cmdline/mod.rs
```diff
@@ -234,24 +221,110 @@ impl Cmdline {
...
-    /// cl.insert_str("foobar");
-    /// let cl_cstring = CString::new(cl).unwrap();
-    /// assert_eq!(cl_cstring.to_str().unwrap(), "foobar");
+    /// let mut cl = Cmdline::new(100).unwrap();
+    /// cl.insert_str("foobar").unwrap();
+    /// assert_eq!(cl.as_cstring().unwrap().as_bytes_with_nul(), b"foobar\0");
     /// ```
     pub fn insert_str<T: AsRef<str>>(&mut self, slug: T) -> Result<()> {
-        let s = slug.as_ref();
+        // Step 1: Check if the string provided is a valid boot arg string and remove any
```
aghecenco, line 229: These two `insert`s are really similar so I think you can comfortably extract the common bits into (yet another) helper - the flow is basically:
1. check that `slug` fits
2. put `slug` in either `boot_args` or `init_args`

So you could compress step 1. into a checker and just pass it how many extra chars it needs to accommodate, something like

```rust
fn check_size<T: AsRef<str>>(&mut self, slug: &T, num_extra_chars: usize) -> Result<()> {
        // Step 1: Check if the string provided is a valid boot arg string and remove any
        // leading or trailing whitespaces.
        let s = slug.as_ref().trim();
        valid_str(s)?;

        // Step 2: Check if cmdline capacity is not exceeded when inserting the arg
        // string provided.
        let cmdline_size = self.get_null_terminated_representation_size()
        .checked_add(num_extra_chars).ok_or(Error::TooLarge)?
        .checked_add(s.len()).ok_or(Error::TooLarge)?;

        if cmdline_size > self.capacity {
            Err(Error::TooLarge)
        } else {
            Ok(())
        }
    }
```

Then call it like this -

```rust
pub fn insert_str<T: AsRef<str>>(&mut self, slug: T) -> Result<()> {
        self.check_size(&slug, if self.boot_args.is_empty() { 0 } else { 1 })?;

        if !self.boot_args.is_empty() {
        // etc
    }

pub fn insert_init_args<T: AsRef<str>>(&mut self, slug: T) -> Result<()> {
        self.check_size(&slug, if self.init_args.is_empty() {
            INIT_ARGS_SEPARATOR.len()
        } else {
            1
        })?;

        if !self.init_args.is_empty() {
        // etc
}
````

File: src/cmdline/mod.rs [hunk 33]
```diff
@@ -343,17 +431,106 @@ impl Cmdline {
...
+        let mut cmdline_size = boot_args.len().checked_add(1).ok_or(Error::TooLarge)?;
+
+        if !init_args.is_empty() {
+            cmdline_size = cmdline_size
+                .checked_add(INIT_ARGS_SEPARATOR.len())
+                .ok_or(Error::TooLarge)?;
+
+            cmdline_size = cmdline_size
+                .checked_add(init_args.len())
+                .ok_or(Error::TooLarge)?;
```
aghecenco, line 507: nit: these `checked_add`s can be chained into something like

```rust
    let cmdline_size = boot_args
            .len()
            .checked_add(1)
            .and_then(|s| {
                if !init_args.is_empty() {
                    s.checked_add(INIT_ARGS_SEPARATOR.len() + init_args.len())
                } else {
                    Some(s)
                }
            })
            .ok_or(Error::TooLarge)?;
```

aghecenco (commented):
I think the only thing missing at this point is the changelog update about the breaking changes, functionality-wise LGTM.

src/cmdline/mod.rs
andreitraistaru: As discussed, I'll let as it is, as long as we still need to have checked_add() for `INIT_ARGS_SEPARATOR.len() + init_args.len()`

src/cmdline/mod.rs
```diff
@@ -223,7 +209,8 @@ impl Cmdline {
         self.insert_str(kv_str)
     }
 
-    /// Validates and inserts a string to the end of the current command line.
+    /// Validates and inserts a string representing boot args
```
andreeaflorescu, line 212: I think this doc comment is a bit misleading. The purpose of the function is to "insert a string in the boot args". And then we can say: "Returns an error if the string is invalid". This was wrong before as well, but since we are updating it, we can also correct it. With the current wording it looks like the function does more than one thing which is not really the case.

src/cmdline/mod.rs
```diff
@@ -343,17 +431,106 @@ impl Cmdline {
...
+        }
+    }
 }
 
 impl From<Cmdline> for Vec<u8> {
     fn from(cmdline: Cmdline) -> Vec<u8> {
-        cmdline.line.into_bytes()
+        match cmdline.as_cstring() {
+            Ok(cmdline_cstring) => cmdline_cstring.into_bytes_with_nul(),
+            Err(_) => Vec::from([0u8]),
```
andreeaflorescu, line 526: This is a silent failure. We should change this to try_from instead and return the error from `into_bytes_with_nul` instead.

andreitraistaru: solved

src/cmdline/mod.rs
```diff
@@ -8,24 +8,21 @@
...
+// this is the minimum required space to accommodate the null terminator and ' -- ' delimiter between boot args and
+// init args
+const CMDLINE_NO_OF_AUXILIARY_CHARS: usize = 5;
+
 /// The error type for command line building operations.
 #[derive(Debug, PartialEq)]
 pub enum Error {
     /// Operation would have resulted in a non-printable ASCII character.
     InvalidAscii,
-    /// Invalid device passed to the kernel command line builder.
```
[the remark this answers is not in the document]
andreitraistaru: done

CHANGELOG.md
```diff
@@ -1,3 +1,18 @@
...
+
+## Added
+- Added `NoBootArgsInserted` error type used when calling `as_cstring` on a `Cmdline`
+  containing only init args.
+- Added `InvalidCapacity` error type used when trying to create a `Cmdline`
+  either with zero capacity (using `new` or `try_from`) or with smaller capacity
+  than required for arguments (using `try_from`).
+- Added `insert_init_args` method allowing insertion of init arguments into `Cmdline`.
+
+## Changed
```
andreeaflorescu, line 11: We also return a `Result` in `new` and check for invalid command line size.

CHANGELOG.md
```diff
@@ -1,3 +1,18 @@
+# [Unreleased]
+
+## Added
+- Added `NoBootArgsInserted` error type used when calling `as_cstring` on a `Cmdline`
+  containing only init args.
+- Added `InvalidCapacity` error type used when trying to create a `Cmdline`
+  either with zero capacity (using `new` or `try_from`) or with smaller capacity
+  than required for arguments (using `try_from`).
+- Added `insert_init_args` method allowing insertion of init arguments into `Cmdline`.
```
andreeaflorescu, line 9: We should also say we added `try_from` that has support for both init args and command line args.

CHANGELOG.md
andreeaflorescu: And we also changed the implementation that converts a command line to a Vec<u8> and added error handling (i.e replaced From<Cmdline> with TryFrom<Cmdline>

andreitraistaru: done

aghecenco (approved):
I am OK with the current changes, the changelog update is maybe a bit too verbose but that can be adjusted when we release.

PR #118, opened 2022-09-19
Changing as_str into as_cstr

andreitraistaru:
Changing as_str() into as_cstr() in order to retrieve a null terminated string from the Cmdline struct.

Signed-off-by: Traistaru Andrei Cristian <atc@amazon.com>

### Summary of the PR

*Please summarize here why the changes in this PR are needed.*

### Requirements

requirements:

  description line.
  test.

### DIFF
--- a/src/cmdline/mod.rs
+++ b/src/cmdline/mod.rs
@@ -8,6 +8,7 @@
 //
 //! Helper for creating valid kernel command line strings.
 
+use std::ffi::CString;
 use std::fmt;
 use std::result;
 
@@ -16,6 +17,8 @@ use vm_memory::{Address, GuestAddress, GuestUsize};
 /// The error type for command line building operations.
 #[derive(Debug, PartialEq, Eq)]
 pub enum Error {
+    /// Null terminator identified in the command line.
+    NullTerminator,
     /// Operation would have resulted in a non-printable ASCII character.
     InvalidAscii,
     /// Invalid device passed to the kernel command line builder.
@@ -35,6 +38,9 @@ pub enum Error {
 impl fmt::Display for Error {
     fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
         match *self {
+            Error::NullTerminator => {
+                write!(f, "Null terminator detected in the command line structure.")
+            }
             Error::InvalidAscii => write!(f, "String contains a non-printable ASCII character."),
             Error::InvalidDevice(ref dev) => write!(
                 f,
@@ -156,11 +162,9 @@ impl Cmdline {
     ///
     /// ```rust
     /// # use linux_loader::cmdline::*;
-    /// # use std::ffi::CString;
     /// let mut cl = Cmdline::new(100);
     /// cl.insert("foo", "bar");
-    /// let cl_cstring = CString::new(cl).unwrap();
-    /// assert_eq!(cl_cstring.to_str().unwrap(), "foo=bar");
+    /// assert_eq!(cl.as_cstring().unwrap().as_bytes_with_nul(), b"foo=bar\0");
     /// ```
     pub fn insert<T: AsRef<str>>(&mut self, key: T, val: T) -> Result<()> {
         let k = key.as_ref();
@@ -248,7 +252,12 @@ impl Cmdline {
         Ok(())
     }
 
-    /// Returns the string representation of the command line without the nul terminator.
+    /// Returns a C compatible representation of the command line
+    /// The Linux kernel expects a null terminated cmdline according to the source:
+    /// https://elixir.bootlin.com/linux/v5.10.139/source/kernel/params.c#L179
+    ///
+    /// To get bytes of the cmdline to be written in guest's memory (including the
+    /// null terminator) from this representation, use CString::as_bytes_with_nul()
     ///
     /// # Examples
     ///
@@ -256,10 +265,10 @@ impl Cmdline {
     /// # use linux_loader::cmdline::*;
     /// let mut cl = Cmdline::new(10);
     /// cl.insert_str("foobar");
-    /// assert_eq!(cl.as_str(), "foobar");
+    /// assert_eq!(cl.as_cstring().unwrap().as_bytes_with_nul(), b"foobar\0");
     /// ```
-    pub fn as_str(&self) -> &str {
-        self.line.as_str()
+    pub fn as_cstring(&self) -> Result<CString> {
+        CString::new(self.line.to_string()).map_err(|_| Error::NullTerminator)
     }
 
     /// Adds a virtio MMIO device to the kernel command line.
@@ -356,9 +365,12 @@ mod tests {
     #[test]
     fn test_insert_hello_world() {
         let mut cl = Cmdline::new(100);
-        assert_eq!(cl.as_str(), "");
+        assert_eq!(cl.as_cstring().unwrap().as_bytes_with_nul(), b"\0");
         assert!(cl.insert("hello", "world").is_ok());
-        assert_eq!(cl.as_str(), "hello=world");
+        assert_eq!(
+            cl.as_cstring().unwrap().as_bytes_with_nul(),
+            b"hello=world\0"
+        );
 
         let s = CString::new(cl).expect("failed to create CString from Cmdline");
         assert_eq!(s, CString::new("hello=world").unwrap());
@@ -369,7 +381,10 @@ mod tests {
         let mut cl = Cmdline::new(100);
         assert!(cl.insert("hello", "world").is_ok());
         assert!(cl.insert("foo", "bar").is_ok());
-        assert_eq!(cl.as_str(), "hello=world foo=bar");
+        assert_eq!(
+            cl.as_cstring().unwrap().as_bytes_with_nul(),
+            b"hello=world foo=bar\0"
+        );
     }
 
     #[test]
@@ -379,7 +394,7 @@ mod tests {
         assert_eq!(cl.insert("a", "b "), Err(Error::HasSpace));
         assert_eq!(cl.insert("a ", "b "), Err(Error::HasSpace));
         assert_eq!(cl.insert(" a", "b"), Err(Error::HasSpace));
-        assert_eq!(cl.as_str(), "");
+        assert_eq!(cl.as_cstring().unwrap().as_bytes_with_nul(), b"\0");
     }
 
     #[test]
@@ -390,7 +405,7 @@ mod tests {
         assert_eq!(cl.insert("a=", "b "), Err(Error::HasEquals));
         assert_eq!(cl.insert("=a", "b"), Err(Error::HasEquals));
         assert_eq!(cl.insert("a", "=b"), Err(Error::HasEquals));
-        assert_eq!(cl.as_str(), "");
+        assert_eq!(cl.as_cstring().unwrap().as_bytes_with_nul(), b"\0");
     }
 
     #[test]
@@ -398,17 +413,20 @@ mod tests {
         let mut cl = Cmdline::new(100);
         assert_eq!(cl.insert("heart", "💖"), Err(Error::InvalidAscii));
         assert_eq!(cl.insert("💖", "love"), Err(Error::InvalidAscii));
-        assert_eq!(cl.as_str(), "");
+        assert_eq!(cl.as_cstring().unwrap().as_bytes_with_nul(), b"\0");
     }
 
     #[test]
     fn test_insert_string() {
         let mut cl = Cmdline::new(13);
-        assert_eq!(cl.as_str(), "");
+        assert_eq!(cl.as_cstring().unwrap().as_bytes_with_nul(), b"\0");
         assert!(cl.insert_str("noapic").is_ok());
-        assert_eq!(cl.as_str(), "noapic");
+        assert_eq!(cl.as_cstring().unwrap().as_bytes_with_nul(), b"noapic\0");
         assert!(cl.insert_str("nopci").is_ok());
-        assert_eq!(cl.as_str(), "noapic nopci");
+        assert_eq!(
+            cl.as_cstring().unwrap().as_bytes_with_nul(),
+            b"noapic nopci\0"
+        );
     }
 
     #[test]
@@ -420,7 +438,7 @@ mod tests {
         assert!(cl.insert("a", "b").is_ok());
         assert_eq!(cl.insert("a", "b"), Err(Error::TooLarge));
         assert_eq!(cl.insert_str("a"), Err(Error::TooLarge));
-        assert_eq!(cl.as_str(), "a=b");
+        assert_eq!(cl.as_cstring().unwrap().as_bytes_with_nul(), b"a=b\0");
 
         let mut cl = Cmdline::new(10);
         assert!(cl.insert("ab", "ba").is_ok()); // adds 5 length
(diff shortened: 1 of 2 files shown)


src/cmdline/mod.rs
```diff
@@ -256,10 +258,13 @@ impl Cmdline {
...
     /// cl.insert_str("foobar");
-    /// assert_eq!(cl.as_str(), "foobar");
+    /// assert_eq!(cl.as_cstr(), "foobar");
     /// ```
-    pub fn as_str(&self) -> &str {
-        self.line.as_str()
+    pub fn as_cstr(&self) -> &Cstr{
+        match CString::new(self.line.to_string()) {
+            Ok(cmdline_cstring) => cmdline_cstring.as_c_str(),
+            Err(_) => CString::new("".to_string()),
```
andreeaflorescu, line 266: The error should be propagated and we can turn this into a `Result`.

src/cmdline/mod.rs
```diff
@@ -256,10 +258,13 @@ impl Cmdline {
     /// # use linux_loader::cmdline::*;
     /// let mut cl = Cmdline::new(10);
     /// cl.insert_str("foobar");
-    /// assert_eq!(cl.as_str(), "foobar");
+    /// assert_eq!(cl.as_cstr(), "foobar");
     /// ```
-    pub fn as_str(&self) -> &str {
-        self.line.as_str()
+    pub fn as_cstr(&self) -> &Cstr{
```
andreeaflorescu, line 263: We should also add a regression test that validates that the string is null terminated.

Can you also add more details to the commit message? We should specify here that this was a bug introduced by #72 and that the kernel is reading from the command line until it finds the null terminator, which is why we need to be sure to add it.

andreitraistaru:
I'll fix the comments and also the test that are currently failing. Also, I'll add a test to validate the null terminator.

src/cmdline/mod.rs
andreitraistaru: I updated the commit message. Also, fixed the tests from pipeline. I introduced various checks that the as_cstring() method returns a null terminated string. I also modified the existing unit test from loader to check for the null terminator. Should I add a separate test in linux loader just to check the null terminated string returned by as_cstring? I don't think is necessary as long as the current tests checks the bytes used for writing the cmdline into guest's memory to contain the null terminator. Let me a comment if you think there are other tests that I can add.

src/loader/mod.rs
```diff
@@ -211,7 +211,7 @@ pub fn load_cmdline<M: GuestMemory>(
     guest_addr: GuestAddress,
     cmdline: &Cmdline,
 ) -> Result<()> {
-    let len = cmdline.as_str().len();
+    let len = cmdline.as_cstring().unwrap().to_bytes_with_nul().len();
```
lauralt, line 214: It's not safe to use unwraps in library code, you should either propagate the error or explain why the unwrap can not fail.
Also I don't understand the check below. If we use `to_bytes_with_nul` then we will have a nul terminator so the length can't be 0 anyway. I think we want to use `to_bytes` here.

src/loader/mod.rs
```diff
@@ -258,9 +261,15 @@ mod tests {
     fn test_cmdline_write_end() {
         let gm = create_guest_mem();
         let mut cmdline_address = GuestAddress(45);
+
         let mut cl = Cmdline::new(10);
-        cl.insert_str("1234").unwrap();
+        cl.insert_str("123").unwrap();
```
lauralt, line 266: Why this change?

src/loader/mod.rs
```diff
@@ -271,7 +280,20 @@ mod tests {
...
         let val: u8 = gm.read_obj(cmdline_address).unwrap();
-        assert_eq!(val, b'4');
+        assert_eq!(val, b'\0');
+
+        // Overwrite guest's memory where cmdline was previously written with a shorter
+        // cmdline to check that the null terminator is written in guest's memory
+        let mut cl = Cmdline::new(10);
+        cl.insert_str("12").unwrap();
+        assert_eq!(cl.as_cstring().unwrap().as_bytes_with_nul(), &[49, 50, 0]);
+        assert_eq!(Ok(()), load_cmdline(&gm, cmdline_address, &cl));
```
lauralt, line 290:
```suggestion
        load_cmdline(&gm, cmdline_address, &cl).unwrap();
```

andreitraistaru:
Thanks for the review. I'm working on addressing the comments now. I'll ping you when done.

src/cmdline/mod.rs
```diff
@@ -356,9 +358,13 @@ mod tests {
     #[test]
     fn test_insert_hello_world() {
         let mut cl = Cmdline::new(100);
-        assert_eq!(cl.as_str(), "");
+        assert_eq!(cl.as_cstring().unwrap().as_bytes_with_nul(), b"\0");
         assert!(cl.insert("hello", "world").is_ok());
-        assert_eq!(cl.as_str(), "hello=world");
+        assert_ne!(cl.as_cstring().unwrap().as_bytes_with_nul(), b"hello=world");
```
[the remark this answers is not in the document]
andreitraistaru: yes.... that's true. I was a litlle bit afraid of not missing the null terminator when converting from one type to another, but as long as we convert into &[u8] it definitely should add a 0 byte at the end as null terminator. Fixing that for all unit tests in the same situation.

src/cmdline/mod.rs
```diff
@@ -369,7 +375,14 @@ mod tests {
         let mut cl = Cmdline::new(100);
         assert!(cl.insert("hello", "world").is_ok());
         assert!(cl.insert("foo", "bar").is_ok());
-        assert_eq!(cl.as_str(), "hello=world foo=bar");
+        assert_ne!(
```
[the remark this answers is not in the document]
andreitraistaru: Done

src/loader/mod.rs
andreitraistaru: The actual change that I wanted to introduce was to catch the overwriting of the cmdline in guest's memory such that the byte of the null terminator (before overwriting to memory) not to be 0. This way I could make sure that the cmdline really contains the null terminator that is written to memory.

The change of the string from "1234" to "123" was just to reduce the number of bytes to check from the memory (and also check the entire cmdline that was written into memory). Definitely, I could have checked just the bytes around the null terminator or maybe use totally different bytes for the overwriting but it was not something more complex than that.

src/loader/mod.rs
andreitraistaru: yes, doing this way; replicating for the first write of the cmdline into memory; if the load_cmdline() fails, the unwrap() will panic so the test fails.

src/loader/mod.rs
andreitraistaru: I change the first cmdline that is written into memory to contain "123" and the second one to contain "45". Doing like this, when calling load_cmdline() with the same starting address in memory, there is no risk to make confusions between which byte was written by which call of the load_cmdline().

src/loader/mod.rs
andreitraistaru: Yes, there is a problem here, but I would solve it by removing the unnecessary check if len is 0 (as you said, there is no possibility to have 0 bytes because we have the null terminator at least). In case of an empty cmdline, I would write the \0 to guest memory to make sure the kernel will stop parsing the cmdline right away - otherwise, we do not have any guarantee that there is a \0 where the kernel expect to find a cmdline.

src/cmdline/mod.rs
```diff
@@ -248,18 +247,21 @@ impl Cmdline {
...
-    /// assert_eq!(cl.as_str(), "foobar");
+    /// assert_eq!(cl.as_cstring().unwrap().as_bytes_with_nul(), b"foobar\0");
     /// ```
-    pub fn as_str(&self) -> &str {
-        self.line.as_str()
+    pub fn as_cstring(&self) -> Result<CString> {
+        match CString::new(self.line.to_string()) {
+            Ok(cmdline_cstring) => Ok(cmdline_cstring),
+            Err(_) => Err(Error::InvalidAscii),
+        }
```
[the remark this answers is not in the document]
andreitraistaru: Done, both adding a new error variant and refactoring in a shorter piece of code as indicated. Thanks!

src/cmdline/mod.rs
```diff
@@ -445,25 +467,45 @@ mod tests {
             .add_virtio_mmio_device(1, GuestAddress(0), 1, None)
             .is_ok());
         let mut expected_str = "virtio_mmio.device=1@0x0:1".to_string();
-        assert_eq!(cl.as_str(), &expected_str);
+        assert_eq!(
+            cl.as_cstring().unwrap().as_bytes_with_nul(),
+            CString::new(expected_str.as_bytes())
+                .unwrap()
+                .as_bytes_with_nul()
+        );
```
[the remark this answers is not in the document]
andreitraistaru: Done

src/cmdline/mod.rs
```diff
@@ -16,6 +17,8 @@ use vm_memory::{Address, GuestAddress, GuestUsize};
 /// The error type for command line building operations.
 #[derive(Debug, PartialEq, Eq)]
 pub enum Error {
+    /// Null terminator identified in the command line.
+    NullTerminatorInCmdline,
```
andreeaflorescu, line 21: nit: we don't need to specify "InCmdline". This Error type is defined for Cmdline, so it's obvious from namespacing that the error refers to that.

src/loader/mod.rs
```diff
@@ -211,20 +215,21 @@ pub fn load_cmdline<M: GuestMemory>(
     guest_addr: GuestAddress,
     cmdline: &Cmdline,
 ) -> Result<()> {
-    let len = cmdline.as_str().len();
-    if len == 0 {
-        return Ok(());
-    }
+    let cmdline_cstring = cmdline
```
andreeaflorescu, line 218: Since we only use the bytes I think we can just keep those bytes around by directly calling here `as_bytes_with_null`. We should add a comment here as well that Linux expects the command line to end with a `\0` which is why we're doing this.

src/loader/mod.rs
```diff
@@ -211,20 +215,22 @@ pub fn load_cmdline<M: GuestMemory>(
     guest_addr: GuestAddress,
     cmdline: &Cmdline,
 ) -> Result<()> {
-    let len = cmdline.as_str().len();
-    if len == 0 {
-        return Ok(());
-    }
+    let cmdline_cstring = match cmdline.as_cstring() {
+        Ok(cmdline_cstring_raw) => cmdline_cstring_raw,
+        Err(_) => return Err(Error::InvalidCommandLine),
```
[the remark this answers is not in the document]
andreeaflorescu: Nvm, I just checked and CString only returns an error for Null char. We don't need to propagate it any longer.

src/cmdline/mod.rs
```diff
@@ -248,18 +252,18 @@ impl Cmdline {
         Ok(())
     }
 
-    /// Returns the string representation of the command line without the nul terminator.
+    /// Returns the representation of the command line containing the null terminator.
```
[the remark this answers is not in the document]
andreitraistaru: Updated the comment.

src/cmdline/mod.rs
```diff
@@ -248,18 +252,24 @@ impl Cmdline {
         Ok(())
     }
 
-    /// Returns the string representation of the command line without the nul terminator.
+    /// Returns a C compatible representation of the command line (null terminated)
+    /// The Linux kernel expects a null terminated cmdline according to the source:
+    /// https://elixir.bootlin.com/linux/v5.10.139/source/kernel/params.c#L179
+    ///
+    /// To get bytes of the cmdline to be written in guest's memory (including the
+    /// null terminator) from this representation, method CString::as_bytes_with_null()
```
andreeaflorescu, line 260: nit: missing the verb.

src/cmdline/mod.rs
```diff
@@ -248,18 +252,24 @@ impl Cmdline {
         Ok(())
     }
 
-    /// Returns the string representation of the command line without the nul terminator.
+    /// Returns a C compatible representation of the command line (null terminated)
```
andreeaflorescu, line 255: I don't think that we should write here null terminated because it's just an abstraction. To get the null terminated bytes you need to call a specific function.

src/loader/mod.rs
```diff
@@ -211,20 +215,25 @@ pub fn load_cmdline<M: GuestMemory>(
     guest_addr: GuestAddress,
     cmdline: &Cmdline,
 ) -> Result<()> {
-    let len = cmdline.as_str().len();
-    if len == 0 {
-        return Ok(());
-    }
+    // We need to get a vector of bytes representing the cmdline
```
andreeaflorescu, line 218: This comment is no longer accurate, maybe we can rephrase it to something like: "We need a null terminated string because that's what the linux kernel expects when parsing the command line".

andreeaflorescu (dismissed):

src/cmdline/mod.rs
```diff
@@ -248,18 +252,23 @@ impl Cmdline {
         Ok(())
     }
 
-    /// Returns the string representation of the command line without the nul terminator.
+    /// Returns a C compatible representation of the command line
+    /// The Linux kernel expects a null terminated cmdline according to the source:
+    /// https://elixir.bootlin.com/linux/v5.10.139/source/kernel/params.c#L179
+    ///
+    /// To get bytes of the cmdline to be written in guest's memory (including the
+    /// null terminator) from this representation, use CString::as_bytes_with_null()
```
lauralt, line 260: nit:
```suggestion
    /// null terminator) from this representation, use CString::as_bytes_with_nul()
```

src/loader/mod.rs
```diff
@@ -211,20 +215,24 @@ pub fn load_cmdline<M: GuestMemory>(
...
+        .as_cstring()
+        .map_err(|_| Error::InvalidCommandLine)?;
+
+    let cmdline_bytes = cmdline_string.as_bytes_with_nul();
 
     let end = guest_addr
-        .checked_add(len as u64 + 1)
-        .ok_or(Error::CommandLineOverflow)?; // Extra for null termination.
+        .checked_add(cmdline_bytes.len() as u64)
+        .ok_or(Error::CommandLineOverflow)?;
```
lauralt, line 229: Hmm I just noticed.. so we are returning an error if the first address after the command line area is outside the u64 range. Shouldn't we actually do that for the last address in the command line region?

lauralt: Hmm, looks like we also don't have a test for this corner case.

lauralt (commented):
You should also squash your commits.

src/loader/mod.rs
andreitraistaru: Regarding the tests we have, there is a test that checks the overflow of the memory. Regarding the overflow of that addition ... yes, we don't have one; I'll add this usecase in the existing overflow tes

andreitraistaru: To be honest, the way we compute the value of the end variable seems to be not accurate; with the current implementation, theoretically, we throw an error even if the the cmdline (including null terminator) is within guest memory's range, but the null terminator is in the last byte of the guest (because the next address will be bigger then what guest_mem.last_addr() returns which is the last byte of the guest's memory) - this is a very specific and improbable usecase... I don't even know if it's technically possible to store the cmdline at the end of the guest memory but in the context of this comment, it is worth mentioning this aspect. Am I right? Did I miss something? Should I change`let end = guest_addr
        .checked_add(cmdline_bytes.len() as u64)
        .ok_or(Error::CommandLineOverflow)?;` into `let end = guest_addr
        .checked_add((cmdline_bytes.len() - 1) as u64)
        .ok_or(Error::CommandLineOverflow)?;` ?

andreitraistaru: Leaving this comment unaddressed for further discussions. I'll push the changes for the other comments.

lauralt: > To be honest, the way we compute the value of the end variable seems to be not accurate; with the current implementation, theoretically, we throw an error even if the the cmdline (including null terminator) is within guest memory's range, but the null terminator is in the last byte of the guest (because the next address will be bigger then what guest_mem.last_addr() returns which is the last byte of the guest's memory) - this is a very specific and improbable usecase... I don't even know if it's technically possible to store the cmdline at the end of the guest memory but in the context of this comment, it is worth mentioning this aspect. Am I right? Did I miss something? Should I change`let end = guest_addr.checked_add(cmdline_bytes.len() as u64).ok_or(Error::CommandLineOverflow)?;` into `let end = guest_addr.checked_add((cmdline_bytes.len() - 1) as u64).ok_or(Error::CommandLineOverflow)?;` ?

Yes, this is exactly what I was suggesting.

andreitraistaru: Ah... reading your comment I understood that we should somehow check if the load_cmdline stores the cmdline bytes into a range of addresses that the linux kernel expects to find the cmdline; sorry for my misunderstanding; fixing that right now

src/loader/mod.rs
```diff
@@ -211,20 +215,24 @@ pub fn load_cmdline<M: GuestMemory>(
...
+    let cmdline_string = cmdline
+        .as_cstring()
+        .map_err(|_| Error::InvalidCommandLine)?;
+
+    let cmdline_bytes = cmdline_string.as_bytes_with_nul();
 
     let end = guest_addr
-        .checked_add(len as u64 + 1)
-        .ok_or(Error::CommandLineOverflow)?; // Extra for null termination.
+        .checked_add((cmdline_bytes.len() - 1) as u64)
```
lauralt, line 228: nit: I would leave a comment here explaining that the subtraction can not underflow because cmdline_bytes has at least one byte (the nul terminator),

lauralt (approved):
:+1:

src/loader/mod.rs
andreitraistaru: adding that as well

lauralt (approved):

andreeaflorescu (approved):

PR #119, opened 2022-09-23
release v0.6.0

andreeaflorescu:
### Summary of the PR

Details in the CHANGELOG.

**NOTE**: once we publish the new version I will yank v0.4.0 and v0.5.0 because of a critical bug in the way we're loading the kernel command line into guest memory. The 2 aforemntioned releases do not write a null terminator at the end of the command line which means that the kernel might end up reading past the intended command line size. This is typically not the case in a common VMM scenario because the Guest Memory is typically zeroed, so we do get a null terminated command line by luck.

### Requirements

requirements:

  description line.
  test.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,20 @@
+# [v0.6.0]
+
+## Changed
+- Crate is now using edition 2021.
+
+## Added
+- Derived `Eq` for `Error` types and the `PvhBootCapability` enum.
+
+## Fixed
+- Fixed a bug in `load_cmdline` due to which the command line was not null
+  terminated. This resulted in a change in the `Cmdline` API where instead of
+  returning the cmdline as a String, we're now returning it as a `CString` as
+  the latter has support for converting it to a null terminated bytes array.
+- Fixed an off-by-one error in load_cmdline, where we were doing validations
+  on the first address after the command line memory region, instead of the
+  last inclusive one of it.
+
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -1,3 +1,17 @@
+# [v0.6.0]
+
+## Changed
+- Crate is now using edition 2021.
+
+## Added
+- Derived `Eq` for `Error` types and the `PvhBootCapability` enum.
+
+## Fixed
+- Fixed a bug in `load_cmdline` due to which the command line was not null
```
lauralt, line 10: We should probably mention about [this bug](https://github.com/rust-vmm/linux-loader/pull/118#discussion_r976388631) as well.

andreeaflorescu: We should really enforce separate commits for different bug fixes, otherwise it's very hard to not miss them in the changelog. Can you help me with formulating a changelog entry? I am still not 100% sure I understood what the problem was there.

lauralt: Fixed an off-by-one error in load_cmdline, where we were doing
validations on the first address after the command line memory region,
instead of the last inclusive one of it.

andreeaflorescu: Thanks @lauralt. I've added it, please have another look.

lauralt (approved):

rbradford (approved):

PR #133, opened 2023-02-21
Using rewind instead of seek to start

ita93:
Using rewind instead of seek to start

### DIFF
--- a/src/loader/x86_64/elf/mod.rs
+++ b/src/loader/x86_64/elf/mod.rs
@@ -212,9 +212,7 @@ impl KernelLoader for Elf {
     where
         F: Read + Seek,
     {
-        kernel_image
-            .seek(SeekFrom::Start(0))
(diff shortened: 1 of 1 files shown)


rbradford (commented):
Thank you for your contribution. I'm not 100% sure if this is an improvement - I imagine that more developers are familiar with `.seek()` than `.rewind()`. I know I would have to look it up in the API docs vs instantly recognising what the `.seek()` was trying to achieve.

However, i'm open to others' thoughts of course.

andreeaflorescu:
@rbradford we changed this in vm-virtio as well from Seek(0) to Rewind because it will yield a clippy warning otherwise. I didn't even know that `rewind` existed before that 😆. I think it is okay to merge this.

andreeaflorescu (approved):

rbradford (approved):
Okay, clippy knows best! :-)

PR #148, opened 2023-05-14
Fixing kernel commandline parameter validation

ita93:
According to linux kernel document, the value of command line parameter can contains spaces (with double quotes) or equals.

### Summary of the PR

*Please summarize here why the changes in this PR are needed.*

### Requirements

requirements:

  description line.
  test.

### DIFF
--- a/src/cmdline/mod.rs
+++ b/src/cmdline/mod.rs
@@ -104,7 +104,7 @@ fn contains_double_quotes(s: &str) -> bool {
     if s.len() < 3 {
         return false;
     }
-    return s.chars().skip(1).take(s.len() - 2).any(|c| c == '"');
+    s.chars().skip(1).take(s.len() - 2).any(|c| c == '"')
 }
 
 fn valid_key(s: &str) -> Result<()> {
--- a/src/cmdline/mod.rs
+++ b/src/cmdline/mod.rs
@@ -37,6 +37,10 @@ pub enum Error {
     MmioSize,
     /// Operation would have made the command line too large.
     TooLarge,
+    /// Double-quotes can be used to protect spaces in values
+    NoQuoteSpace,
+    /// A double quote that is in the middle of the value
+    InvalidQuote,
 }
 
 impl fmt::Display for Error {
@@ -56,6 +60,11 @@ impl fmt::Display for Error {
                 "0-sized virtio MMIO device passed to the kernel command line builder."
             ),
             Error::TooLarge => write!(f, "Inserting string would make command line too long."),
+            Error::NoQuoteSpace => write!(
+                f,
+                "Value that contains spaces need to be surrounded by quotes"
+            ),
+            Error::InvalidQuote => write!(f, "Double quote can not be in the middle of the value"),
         }
     }
 }
@@ -79,13 +88,44 @@ fn valid_str(s: &str) -> Result<()> {
     }
 }
 
-fn valid_element(s: &str) -> Result<()> {
+fn is_quoted(s: &str) -> bool {
+    if s.len() < 2 {
+        return false;
+    }
+    if let Some(first_char) = s.chars().next() {
+        if let Some(last_char) = s.chars().last() {
+            return first_char == '"' && last_char == '"';
+        }
+    }
+    false
+}
+
+fn contains_double_quotes(s: &str) -> bool {
+    if s.len() < 3 {
+        return false;
+    }
+    return s.chars().skip(1).take(s.len() - 2).any(|c| c == '"');
+}
+
+fn valid_key(s: &str) -> Result<()> {
     if !s.chars().all(valid_char) {
         Err(Error::InvalidAscii)
     } else if s.contains(' ') {
         Err(Error::HasSpace)
     } else if s.contains('=') {
         Err(Error::HasEquals)
+    } else {
+        Ok(())
+    }
+}
+
+fn valid_value(s: &str) -> Result<()> {
+    if !s.chars().all(valid_char) {
+        Err(Error::InvalidAscii)
+    } else if contains_double_quotes(s) {
+        Err(Error::InvalidQuote)
+    } else if s.contains(' ') && !is_quoted(s) {
+        Err(Error::NoQuoteSpace)
     } else {
         Ok(())
     }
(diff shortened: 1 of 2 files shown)


roypat:
Heyo, thanks for the contribution. Just a few questions, mainly for my own understanding since I couldn't find the answer in the linux docs >.>

What would happen if the value passed looks like `"hello" world"`, e.g. with a spurious quote in the middle? Do we need to also error out in that case? Also, are single quotes allowed for the purpose of preserving spaces? 

From a usability perspective, would it make sense to instead detect if the value the user passes us contains spaces, and add the quotes for them (with maybe some escaping logic for quotes already in the string, although I'm not sure the linux command line allows for escaping them).

ita93:
> Heyo, thanks for the contribution. Just a few questions, mainly for my own understanding since I couldn't find the answer in the linux docs >.>
> 
> What would happen if the value passed looks like `"hello" world"`, e.g. with a spurious quote in the middle? Do we need to also error out in that case? Also, are single quotes allowed for the purpose of preserving spaces?
> 
Only double quotes are allowed. You can't escape " in the value,  your example string is invalid.
> From a usability perspective, would it make sense to instead detect if the value the user passes us contains spaces, and add the quotes for them (with maybe some escaping logic for quotes already in the string, although I'm not sure the linux command line allows for escaping them).
Maybe a good idea.

andreitraistaru:
> > > Heyo, thanks for the contribution. Just a few questions, mainly for my own understanding since I couldn't find the answer in the linux docs >.>
> > > What would happen if the value passed looks like `"hello" world"`, e.g. with a spurious quote in the middle? Do we need to also error out in that case? Also, are single quotes allowed for the purpose of preserving spaces?
> > 
> > 
> > Only double quotes are allowed. You can't escape " in the value, your example string is invalid.
> 
> I see, thanks for explaining! In that case, we should also validate that no quotes are contained in the middle of values, e.g. `if value[1..value.len() - 1].contains('"') {...}`
> 
> > > From a usability perspective, would it make sense to instead detect if the value the user passes us contains spaces, and add the quotes for them (with maybe some escaping logic for quotes already in the string, although I'm not sure the linux command line allows for escaping them).
> > 
> > 
> > Maybe a good idea.
> 
> If we don't have to deal with any complex escaping logic, I think we should do this. Unless I'm missing something, it'd just be
> 
>     1. reject strings that contain quotes anywhere but the beginning or end
> 
>     2. reject strings with unbalanced quotes
> 
>     3. check if string contains spaces.
>        3.1. If it does, and it is unquoted, add quotes, otherwise leaves as is
> 
> 
> What do others think?

Thanks @ita93 and @roypat for getting involved into this!

I do agree with the first 2 points (for point 2, we even have a function [here](https://github.com/rust-vmm/linux-loader/pull/148/files#diff-bdccac30338fe0bd6941e140560075103e077b06d680b8af4044cff2c7a13d67R457) that could help us with the implementation).
For point 3, I would agree to do the checking but I'm not quite sure if it would be the right approach to alter the value of the parameter in any way. I mean, if we get a wrong value for a kernel param, I do not consider the Cmdline's responsibility to correct it and store in the Cmdline as it should. For point 3, I would still reject the insertion and that's it. What do you think?

Also, what happens if we receive a parameter's value like `"foo1  bar1" "foo2 bar3"`? Should we also reject such a string (the case when we get nested double quotes that are balanced)? Not sure from the linux docs, but to me it looks like an invalid value as well...

andreeaflorescu:
There is no activity on this PR for some time, @ita93 are you still interested in getting this merged? If not, can you please close the PR?

ita93:
> There is no activity on this PR for some time, @ita93 are you still interested in getting this merged? If not, can you please close the PR?
Hi
Sorry for the late reply, I'm still interested in this one. I will try posting it in slack to ask for reviewing

ita93:
> There is no activity on this PR for some time, @ita93 are you still interested in getting this merged? If not, can you please close the PR?

Sorry for the late reply, I'm still interested in this one. I will try posting it in Slack to ask for a reviewing

roypat:
(replying now since I saw you rebased this recently, sorry for dropping the ball last year!)

Going over the old discussion from last year, I think the only thing left here is to handle the `"hello "world"` and  `"foo1 bar1" "foo2 bar3"` examples above, and also reject them, right? Also, this should probably get a Changelog entry

rbradford:
You should also rebase your commits to drop the merges.

roypat:
> You should also rebase your commits to drop the merges.

I think they're all merge commits created by the github UI (they're all empty), so they'll be dropped if we merge by rebasing

ita93:
> Going over the old discussion from last year, I think the only thing left here is to handle the `"hello "world"` and `"foo1 bar1" "foo2 bar3"` examples above, and also reject them, right? Also, this should probably get a Changelog entry
I've added a changelog entry in the newest commit.
In summary, the rules will be:
- reject strings that contain quotes anywhere but the beginning or end
- reject strings that contain spaces but are not quoted.

src/cmdline/mod.rs
```diff
@@ -79,7 +85,16 @@ fn valid_str(s: &str) -> Result<()> {
...
 
-fn valid_element(s: &str) -> Result<()> {
+fn is_quoted(s: &str) -> bool {
+    if let Some(first_char) = s.chars().next() {
+        if let Some(last_char) = s.chars().last() {
+            return first_char == '"' && last_char == '"';
+        }
+    }
+    false
+}
```
TimePrinciple, line 98: IMHO this implementation would accept input with only a '"' character, should we check if `next()` and `last()` actually pointing different location 🤔, or else we check the length of `s` which should be at least 2 characters long

TimePrinciple (commented):
Some thoughts :)

roypat (dismissed):
Thanks! Could you squash the last commit such that we don't have commits that modify code introduces in earlier commits though?

roypat (dismissed):

src/cmdline/mod.rs
ita93: Hi, I've updated this function as per your suggestion. Can you help me to review it again, please?

rbradford (dismissed):
Rubber stamping after @roypat's review.

ita93:
Hi @andreeaflorescu, all checks have passed, could you have a look to see if it can be merged? thank you

roypat:
@ita93 seems like clippy is complaining about something - could you fix that up? I think then we're good to merge this, no one else seems to want to weigh in on it :)

roypat (approved):

ShadowCurse (approved):

PR #150, opened 2023-05-15
Update bzimage kernel image download path

### LINKED ISSUE #149: kernel-image-4.9.0 which used in .buildkite test is no longer downloadable
It seems the file was removed from ftp server, 
The requested URL /debian/pool/main/l/linux/kernel-image-4.9.0-13-amd64-di_4.9.228-1_amd64.udeb was not found on this server.

aghecenco:
### Summary of the PR

The integ tests used to download a 4.9 kernel image from the Debian servers, which has been since taken down. Updating to a 5.10 image and adjusting the tests accordingly.

Fixes #149

### Requirements

requirements:

  description line.
  test.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -25,7 +25,7 @@ pe = []
 vm-memory = "0.11.0"
 
 [dev-dependencies]
-criterion = "0.3.5"
+criterion = { version = "0.3.5", features = ["html_reports"] }
 vm-memory = { version = "0.11.0", features = ["backend-mmap"] }
 
 [[bench]]
--- a/.buildkite/download_resources.sh
+++ b/.buildkite/download_resources.sh
@@ -2,13 +2,13 @@
 
 set -e
 
-DEB_NAME="kernel-image-4.9.0-13-amd64-di_4.9.228-1_amd64.udeb"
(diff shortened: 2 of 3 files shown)


Cargo.toml
```diff
@@ -25,7 +25,7 @@ pe = []
 vm-memory = "0.11.0"
 
 [dev-dependencies]
-criterion = "0.3.5"
+criterion = { version = "0.3.5", features = ["html_reports"] }
```
lauralt, line 28: Hmm, the CI is still failing because of some warning.

roypat:
Alright, so I had a quick look into why the CI is failing, and its due to how we collect criterion data. Since we don't check benchmark data into the repository, we run criterion twice: Once against main, and once on the actual PR. However, when checking out main, we try to download the now-removed ubuntu image, causing the failures. I don't think we have any alternative to just force-merging this PR :/

andreeaflorescu:
> Alright, so I had a quick look into why the CI is failing, and its due to how we collect criterion data. Since we don't check benchmark data into the repository, we run criterion twice: Once against main, and once on the actual PR. However, when checking out main, we try to download the now-removed ubuntu image, causing the failures. I don't think we have any alternative to just force-merging this PR :/

Right, this happened in the past as well, so we did force merges. Let's to this again.

PR #176, opened 2024-01-26
Add support for empty init arguments in the command line parser

mschlumpp:
### Summary of the PR

For automated scripts it is easier to keep the trailing `--` separator than to separately check whether it is strictly necessary.
Without this change the crate would include the trailing `--` as a kernel argument. Downstream consumers such as firecracker then proceed to append further parameters to that string (mmio parameters). The final re-assembled string looked similar to this: `<kernel params> -- <mmio-args>`, therefore passing the mmio parameters not to the kernel but to the init part.

### Requirements

requirements:

  test.

### DIFF
--- a/src/cmdline/mod.rs
+++ b/src/cmdline/mod.rs
@@ -832,6 +832,16 @@ mod tests {
 
         assert_eq!(cl.boot_args, "foo=\"bar--baz\" foo");
         assert_eq!(cl.init_args, "");
+
+        let cl = Cmdline::try_from("--foo --bar -- ", CMDLINE_MAX_SIZE).unwrap();
+
+        assert_eq!(cl.boot_args, "--foo --bar");
+        assert_eq!(cl.init_args, "");
+
+        let cl = Cmdline::try_from("--foo --bar --", CMDLINE_MAX_SIZE).unwrap();
+
+        assert_eq!(cl.boot_args, "--foo --bar");
+        assert_eq!(cl.init_args, "");
     }
 
     #[test]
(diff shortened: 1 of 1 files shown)


roypat:
Hi @mschlumpp, 
Thank you for your PR! I think what you're trying to achieve is already partially supported by linux-loader (e.g. the first of the two test cases you add passes even without your changes). The only missing piece is that if the string passed to `try_from` has the form `<boot args> --` (e.g. no trailing space after the `--`), then the ` --` will be considered part of the boot args instead of being stripped. For this, I think we can achieve what you want with a simpler change:

```diff
diff --git a/src/cmdline/mod.rs b/src/cmdline/mod.rs
index 2ac7c31..5687ca3 100644
--- a/src/cmdline/mod.rs
+++ b/src/cmdline/mod.rs
@@ -487,6 +487,10 @@ impl Cmdline {
             ),
         };
 
+        if boot_args.ends_with(INIT_ARGS_SEPARATOR.trim_end()) {
+            boot_args = &boot_args[..boot_args.len() - INIT_ARGS_SEPARATOR.trim().len()];
+        }
+
         boot_args = boot_args.trim();
         init_args = init_args.trim();
 ```

Given that we already trim the trailing ` --` if its followed by a space, that seems like a reasonable change to make, but I'd like to get some input from someone who knows more about the intricacies here (maybe @rbradford?).

Other than that, we should also adjust the documentation of `try_from`, to point out that it strips trailing `--`s. It should also get a changelog entry.

stefano-garzarella:
@jschlumpp are you still interested in getting this PR merged? If yes, could you please re-open it in https://github.com/rust-vmm/rust-vmm/. We moved `linux-loader` there

PR #179, opened 2024-05-11
configurator: Load hvm_modlist_entry into guest memory when requested.

nichtverstehen:
### Summary of the PR

Previously module descriptions in BootParams were silently ignored. This change loads them according to the PVH boot spec.

### Requirements

requirements:

  test.


CHANGELOG.md
```diff
@@ -1,3 +1,9 @@
+# Upcoming Release
+
+## Changed
+
+- [[#177](https://github.com/rust-vmm/linux-loader/pull/176)] Implemented loading of PVH module records.
```
roypat, line 5: Mh, I think this should either be "Added" (if this was intentionally not implemented in the past) or "Fixed" (if it was an oversight).

src/configurator/x86_64/pvh.rs
```diff
@@ -164,6 +173,16 @@ impl BootConfigurator for PvhBootConfigurator {
...
             .map_err(|_| Error::StartInfoSetup)?;
 
+        if let Some(modules) = params.modules.as_ref() {
+            let modules_addr = params.modules_start.ok_or(Error::ModulesAddressMissing)?;
+            guest_memory
+                .checked_offset(modules_addr, modules.len())
+                .ok_or(Error::ModulesPastRamEnd)?;
+            guest_memory
+                .write_slice(modules.as_slice(), modules_addr)
+                .map_err(|_| Error::ModulesSetup)?;
```
roypat, line 183: Mh, the first check technically only verifies that `modules_addr + modules.len()` is a valid guest address, it doesn't check anything about range `[modules_addr, module_addr + modules.len())`. However, `write_slice` already does verify the entire range, so if we do not particularly care about having a separate error variant, we can just drop the `checked_offset` check. Otherwise, we can do 
 ```suggestion
            let slice = guest_memory
                 .get_slice(modules_addr, modules.len())
                 .ok_or(Error::ModulesPastRamEnd)?;
             slice
                 .write_slice(modules.as_slice(), modules_addr)
                 .map_err(|_| Error::ModulesSetup)?;
```

roypat (commented):
Thanks for this PR. I've tested it locally with Firecracker's PVH branch and can confirm that modules indeed now work (e.g. with this PR, I can boot an initrd via PVH in Firecracker)

CHANGELOG.md
nichtverstehen: Thanks for the review! :) Updated the changeling entry (also expanded to make it less obscure)

src/configurator/x86_64/pvh.rs
nichtverstehen: Good point! Removed the extra check.

roypat (approved):

JonathanWoollett-Light (approved):

PR #163, opened 2023-09-28
loader: riscv: Initial support

endeneer:
Initial porting to support loading Linux PE Image and dtb on RISC-V platform.

### Summary of the PR

Hi @roypat, as mentioned in https://github.com/rust-vmm/kvm-bindings/pull/92#issuecomment-1737129242, here is the commit porting `linux-loader` to RISC-V.

#### Test result
```
running 20 tests
test cmdline::tests::test_error_try_from ... ok$<2>
test cmdline::tests::test_add_virtio_mmio_device ... ok$<2>
test cmdline::tests::test_as_cstring ... ok$<2>
test cmdline::tests::test_insert_emoji ... ok$<2>
test cmdline::tests::test_insert_equals ... ok$<2>
test cmdline::tests::test_insert_hello_world ... ok$<2>
test cmdline::tests::test_insert_multi ... ok$<2>
test cmdline::tests::test_insert_space ... ok$<2>
test cmdline::tests::test_insert_string ... ok$<2>
test cmdline::tests::test_insert_kv ... ok$<2>
test cmdline::tests::test_partial_eq ... ok$<2>
test cmdline::tests::test_try_from_cmdline_for_vec ... ok$<2>
test cmdline::tests::test_try_from ... ok$<2>
test cmdline::tests::test_insert_too_large ... ok$<2>
test configurator::tests::test_error_messages ... ok$<2>
test configurator::tests::test_bootparam_list_addition ... ok$<2>
test configurator::tests::test_bootparams ... ok$<2>
test loader::tests::test_cmdline_overflow ... ok$<2>
test loader::riscv::pe::tests::load_image ... ok$<2>
test loader::tests::test_cmdline_write_end_regresion ... ok$<2>

test result: ok$<2>. 20 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.09s
```

### Requirements

Before submitting your PR, please make sure you addressed the following
requirements:

- [x] All commits in this PR are signed (with `git commit -s`), and the commit
  message has max 60 characters for the summary and max 75 characters for each
  description line.
- [x] All added/changed functionality has a corresponding unit/integration
  test.
- [N/A] All added/changed public-facing functionality has entries in the "Upcoming 
  Release" section of CHANGELOG.md (if no such section exists, please create one).
- [x] Any newly added `unsafe` code is properly documented.

PR #190, opened 2024-09-11
Introduce RISC-V architecture support

TimePrinciple:
### Summary of the PR

Introduce RISC-V architecture to `loader`, bringing #163 forward.

### Requirements

requirements:

  test.


TimePrinciple:
@roypat @rbradford @ShadowCurse  PTAL when convenient :)

rbradford (dismissed):

benches/riscv64/mod.rs
roypat, line 1: Instead of duplicating this entire benchmark, let's just rename the `aarch64.rs` file to something more appropriate (`fdt.rs`, maybe), and simply have it cfg'd with `any(target_arch = "aarch64", target_arch = "riscv64")

src/configurator/riscv64/fdt.rs
roypat, line 1: This file is also a 1:1 copy of the one in the aarch64 module. Let's figure out a way to reuse these somehow

src/loader/riscv64/pe/mod.rs
```diff
@@ -0,0 +1,238 @@
...
+    /// # Arguments
+    ///
+    /// * `guest_mem` - The guest memory where the kernel image is loaded.
+    /// * `kernel_offset` - 2MB-aligned base addres in guest memory at which to load the kernel.
+    /// * `kernel_image` - Input Image format kernel image.
+    /// * `highmem_start_address` - ignored on RISC-V.
+    ///
+    /// # Returns
+    /// * KernelLoaderResult
+    fn load<F, M: GuestMemory>(
```
roypat, line 108: This is also very similar to the existing aarch64 code. Can we reuse part of it somehow?

roypat (changes requested):
I feel like there is a lot of code duplicated between the ARM and RISC implementations that doesn't really need to be duplicated like this. Can you have a look at that, to see where we can reuse instead? :o

TimePrinciple:
> I feel like there is a lot of code duplicated between the ARM and RISC implementations that doesn't really need to be duplicated like this. Can you have a look at that, to see where we can reuse instead? :o

Yes, I noticed that, I just copied these files incase there is something to be diverged in the future

I can try to merge the common part as possible, but what should we name that, like `risc`s ? :o

rbradford:
> > I feel like there is a lot of code duplicated between the ARM and RISC implementations that doesn't really need to be duplicated like this. Can you have a look at that, to see where we can reuse instead? :o
> 
> Yes, I noticed that, I just copied these files incase there is something to be diverged in the future
> 
> I can try to merge the common part as possible, but what should we name that, like `risc`s ? :o

Maybe move he fdt code can be moved to a higher level and then `pub use` to expose it from the arch module and keep the same API?

rbradford (approved):
After the refactoring this change is lovely and compact.

roypat (approved):

PR #191, opened 2024-09-26
Print inner errors

ShadowCurse:
### Summary of the PR

Print inner errors in `Display` impl of errors.

### Requirements

requirements:

- [ ] All commits in this PR have Signed-Off-By trailers (with
  test.


src/loader/mod.rs
```diff
@@ -75,22 +75,21 @@ pub type Result<T> = std::result::Result<T, Error>;
 
 impl fmt::Display for Error {
     fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
-        let desc = match self {
+        write!(f, "Kernel Loader: ")?;
+        match self {
             #[cfg(all(feature = "bzimage", any(target_arch = "x86", target_arch = "x86_64")))]
-            Error::Bzimage(ref _e) => "failed to load bzImage kernel image",
+            Error::Bzimage(ref e) => write!(f, "failed to load bzImage kernel image: {e}"),
```
TimePrinciple, line 81:
```suggestion
            Error::Bzimage(ref e) => format!("failed to load bzImage kernel image: {e}"),
```

TimePrinciple: How about use `format!` in the `match` block, and call `write!` once to print the inner message. Would calling `write!` twice lead to performance regression :)

ShadowCurse: The issue with `format!` is that we will need to make temporary strings (e.g. do allocations) for all error cases. I can add the `Kernel Loader: ` part of the first `write` into all writes so there will be only 1 `write!` call. But in any case, this is an error case, so I doubt the performance is a big concern here.

TimePrinciple: Silly of me, I recollected these "write"s previously when dealing with write operation into `fifo`s, and indeed this is in an error case.

LGTM :)

roypat (approved):
Maybe we can also use `thiserror` here at some point

rbradford (approved):

PR #205, opened 2025-07-07
loader: Make arm64_image_header public

AlexOrozco1256:
### Summary of the PR

This header is equivalent to the boot_params header found at the top of x86 kernels. We want to make this struct public to use it in a [fw_cfg implementation](https://github.com/cloud-hypervisor/cloud-hypervisor/pull/7117#discussion_r2165060590) that reads the kernel header for both arm and x86

### DIFF
--- a/src/loader/pe/mod.rs
+++ b/src/loader/pe/mod.rs
@@ -73,19 +73,20 @@ impl std::error::Error for Error {}
 #[cfg(target_arch = "aarch64")]
 #[repr(C)]
 #[derive(Debug, Copy, Clone, Default)]
+#[allow(missing_docs)]
 // See kernel doc Documentation/arm64/booting.txt for more information.
 // All these fields should be little endian.
-struct arm64_image_header {
-    code0: u32,
-    code1: u32,
-    text_offset: u64,
-    image_size: u64,
-    flags: u64,
-    res2: u64,
-    res3: u64,
-    res4: u64,
-    magic: u32,
-    res5: u32,
+pub struct arm64_image_header {
+    pub code0: u32,
+    pub code1: u32,
+    pub text_offset: u64,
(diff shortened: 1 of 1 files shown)


likebreath (approved):

likebreath (commented):
The CI is complaining about missing docs for public data structure:

```
error: missing documentation for a struct
--
  | --> src/loader/pe/mod.rs:78:1
  | \|
  | 78 \| pub struct arm64_image_header {
  | \| ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  | \|
  | note: the lint level is defined here
  | --> src/lib.rs:10:9
  | \|
  | 10 \| #![deny(missing_docs)]
  | \|         ^^^^^^^^^^^^

```

AlexOrozco1256:
> The CI is complaining about missing docs for public data structure:
> 
> ```
> error: missing documentation for a struct
> --
>   | --> src/loader/pe/mod.rs:78:1
>   | \|
>   | 78 \| pub struct arm64_image_header {
>   | \| ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
>   | \|
>   | note: the lint level is defined here
>   | --> src/lib.rs:10:9
>   | \|
>   | 10 \| #![deny(missing_docs)]
>   | \|         ^^^^^^^^^^^^
> ```

fixed! thanks for the review!

roypat (commented):
Will the fields also need to be public, or is the struct + its `ByteValued` impl enough for what you want to do? And if we're at it, let's also update the riscv struct :)

likebreath:
@roypat @rbradford Can any of you merge the PR? We will also likely need a new release from that Cloud Hypervisor can properly consume such change. The last release was a while ago too. Please let me know if you need help here. Thank you.

PR #230, opened 2026-03-26
Disable vm-memory default features

germag:
### Summary of the PR

By default vm-memory enables the "rawfd" feature, which causes the compilation to fail on windows.

### Requirements

  test.

### DIFF
--- a/src/loader_gen/x86_64/bootparam.rs
+++ b/src/loader_gen/x86_64/bootparam.rs
@@ -218,6 +218,8 @@ pub struct __kernel_fd_set {
     pub fds_bits: [::std::os::raw::c_ulong; 16usize],
 }
 #[test]
+// On Windows the unsigned long size is 4 bytes, instead of 8
+#[cfg(target_family = "unix")]
 fn bindgen_test_layout___kernel_fd_set() {
     const UNINIT: ::std::mem::MaybeUninit<__kernel_fd_set> = ::std::mem::MaybeUninit::uninit();
     let ptr = UNINIT.as_ptr();
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -26,11 +26,11 @@ pe = ["elf"]
 elf = []
 
 [dependencies]
-vm-memory = ">=0.16.0, <=0.17.1"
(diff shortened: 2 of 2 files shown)


src/loader_gen/x86_64/bootparam.rs
```diff
@@ -218,6 +218,8 @@ pub struct __kernel_fd_set {
     pub fds_bits: [::std::os::raw::c_ulong; 16usize],
 }
 #[test]
+// On Windows the unsigned long size is 4 bytes, instead of 8
+#[cfg(target_family = "unix")]
 fn bindgen_test_layout___kernel_fd_set() {
```
germag: I'm not sure this is correct, since the `fds_bits`'s size will be incorrect, should I change `fds_bits` to `[::std::os::raw::c_ulonglong; 16usize]`  on Windows?

germag: @slp

src/loader_gen/x86_64/bootparam.rs
germag: Same with these, should be changed to `c_ulonglong` ?

lstocchi: I guess so. I think it (libkrun) works fine on windows as the boot struct already uses the longlong through __u64
https://github.com/rust-vmm/linux-loader/blob/83c13bef18fdd8c9de6df47ff7463884cd402a09/src/loader_gen/x86_64/bootparam.rs#L214

but we would have issues if something uses this

src/loader_gen/x86_64/bootparam.rs
lstocchi: I don't think we use it directly on libkrun but this would cause issues on Windows. So it should be updated as you said

slp:
@germag Please rebase this one (we only need the Cargo.toml change now).

germag:
> @germag Please rebase this one (we only need the Cargo.toml change now).

Done

andreeaflorescu (approved):

stefano-garzarella:
@germag are you still interested in getting this PR merged? If yes, could you please re-open it in https://github.com/rust-vmm/rust-vmm/. We moved `linux-loader` there

PR #234, opened 2026-05-15
Add ELF loading support for aarch64 and riscv64

Gelbpunkt:
### Summary of the PR

This adds support for loading ELF files on aarch64 and riscv64. The reason why I'm adding it is because we want to be able to boot [Hermit](https://hermit-os.org/) unikernel images (which are ELF files) without having to wrap them in the Linux kernel `Image` format on these architectures. Booting kernels in ELF format isn't completely unheard of, e.g. U-Boot supports it via the [`bootelf`](https://docs.u-boot.org/en/stable/usage/cmd/bootelf.html) command and QEMU can boot [both ELF and Linux Image files](https://www.qemu.org/docs/master/system/arm/virt.html#hardware-configuration-information-for-bare-metal-programming) via `-kernel` on ARM and RISC-V. Linux' `vmlinux` on aarch64 will not work, they are not meant to be the final bootable kernel executable.

The example testcase I added is the [qemu_el1](https://github.com/google/aarch64-rt/blob/main/examples/qemu_el1.rs) example from `aarch64-rt`.

aarch64 and riscv64 lack PVH support, so the code for that needs to be conditionally compiled out, but that's very straightforward.

Since aarch64 can technically also be big-endian, I also extended the code to not check for little endian, but rather native endian. I have however not run the testsuite on aarch64_be, only on aarch64. I only compile-tested aarch64_be.

The testing was done with a slightly modified version of cloud-hypervisor on aarch64. I don't have access to riscv64 hardware with the hypervisor extension.

### Requirements

requirements:

  test.


Gelbpunkt:
Hi, would it be possible to get some eyes on this? I'd be happy to address any review feedback :)

andreeaflorescu:
Sorry @Gelbpunkt, I've missed this PR. I'll take a look.

One thing to note is that we're currently focusing on moving the rust-vmm crates to the mono-repo. You might need to recreate your PR there if we manage to move linux-loader before your PR gets merged.

src/loader/elf/mod.rs
```diff
@@ -463,16 +463,18 @@ mod tests {
...
 
-    fn make_elf_bin() -> Vec<u8> {
-        let mut v = Vec::new();
-        v.extend_from_slice(include_bytes!("test_elf.bin"));
-        v
+    fn make_x86_64_elf_bin() -> Vec<u8> {
+        include_bytes!("test_x86-64_elf.bin").to_vec()
+    }
+
+    fn make_aarch64_elf_bin() -> Vec<u8> {
```
andreeaflorescu, line 476: This file is significantly larger than the other bin files we have for testing (i.e. 107KB as opposed to less then 1K for the other ones). Why is that the case?

stefano-garzarella:
@Gelbpunkt are you still interested in getting this PR merged? If yes, could you please re-open it in https://github.com/rust-vmm/rust-vmm/. We moved `linux-loader` there

PR #237, opened 2026-06-05
Replace bindgen files with manual translations

slp:
### Summary of the PR

Bindings generated by bindgen are not guaranteed to be cross-platform. In fact, bootparam.rs is known to not be compatible with Windows targets.

The recommended approach is to generate the bindings with bindgen on build, so they get generated on each target platform. But his approach requires clang development packages to be present in the build system, something that some users may find too cumbersome.

Instead, let's replace the bindgen files with manually generated translations from the C headers, using exclusively unambiguous Rust types. Each file also has compile-time checks to ensure the offsets match the expected values.

The structs in these files rarely change, so the maintenance burden of not using bindgen is fairly minimal.

Assisted-by: Claude Code:claude-opus-4.6

### Requirements

requirements:

  test.


rbradford (approved):

slp:
Gentle ping @sameo @roypat @andreitraistaru @ShadowCurse

ShadowCurse (approved):

